PR #13, opened 2017-10-26
SDK: Masks

marcosoliveira91:
Related to: https://github.com/ripe-tech/ripe-core/issues/2290


src/js/visual/config.part.js
```diff
@@ -588,3 +791,30 @@ ripe.Config.prototype._parseDrag = function() {
...
+
+    var maskContext = mask.getContext("2d");
+    var pixel = maskContext.getImageData(x, y, 1, 1);
+    var r = pixel.data[0];
+    var index = parseInt(r);
+
+    return index;
+};
+
+ripe.Config.prototype._fixEvent = function(event) {
```
acaldas: move this to utils and change to fixEvent

src/js/visual/config.part.js
```diff
@@ -79,7 +97,7 @@ ripe.Config.prototype.update = function(state, options) {
     // if the parts and the position haven't changed
     // since the last frame load then ignores the
     // load request and returns immediately
-    var size = this.element.getAttribute("data-current-size");
+    var size = this.element.dataset.current_size;
```
hugo-gomes: the convention is to still use camel case when accessing `dataset` attributes, or am I missing some gotcha for a particular browser?

src/js/visual/config.part.js
```diff
@@ -588,3 +791,18 @@ ripe.Config.prototype._parseDrag = function() {
     var nextFrame = ripe.getFrameKey(nextView, nextPosition);
     this.changeFrame(nextFrame);
 };
+
+ripe.Config.prototype._getCanvasIndex = function(canvas, x, y) {
+    var canvasRealWidth = canvas.getBoundingClientRect().width;
+    var mask = this.element.querySelector(".mask");
+    var ratio = mask.width / canvasRealWidth;
```
hugo-gomes: let's pray `canvasRealWidth` not to be zero :)

src/js/visual/config.part.js
marcosoliveira91: The thing is internally we often use the nomenclature 'data-something_else'.
Previously, on the code, we had all data-attribute names separated by hyphen ('data-current-size') so we updated them to the hyphen/underscore version ('data-current_size'). 

Should we change it?

marcosoliveira91: By the way, when we perform "element.dataset.currenSize" the result will be "data-current-size".

joamag: Just as a side note I was wrong and the `-` character is allowed in the data attributes, but let's just keep using the `_` character instead https://stackoverflow.com/questions/33068777/what-characters-are-allowed-not-allowed-in-the-names-of-custom-data-attribut

joamag:
@marcosoliveira91 please fix the merge conflicts

PR #24, opened 2018-02-02
fixed highlight bug

acaldas:
https://github.com/ripe-tech/ripe-core/issues/2870

Fixed a bug when programatically highlighting a part that is already loaded.

### DIFF
--- a/src/js/visual/configurator.js
+++ b/src/js/visual/configurator.js
@@ -21,8 +21,10 @@ ripe.Configurator.prototype.init = function() {
     this.interval = this.options.interval || 0;
     this.maskOpacity = this.options.maskOpacity || 0.4;
     this.maskDuration = this.options.maskDuration || 150;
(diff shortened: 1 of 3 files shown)


hugo-gomes:
@acaldas create an [new bug issue](https://github.com/ripe-tech/ripe-core/issues/new?labels=triage&template=bug.md) please

joamag:
@acaldas let's be careful with the indentation to avoid unwanted diffs ... :)

src/js/visual/configurator.js
```diff
@@ -276,6 +276,7 @@ ripe.Configurator.prototype.highlight = function(part, options) {
     var frontMask = this.element.querySelector(".front-mask");
     var src = frontMask.getAttribute("src");
     if (src === url) {
+        this.element.classList.add("highlight");
```
joamag: @acaldas I think it's better to move the highlight operation (at the end of the function) up in the function (before this return) that way you only have one highlight add class operation

PR #26, opened 2018-02-12
added initials support

acaldas:
https://github.com/ripe-tech/ripe-core/issues/2913


src/js/base/ripe.js
```diff
@@ -20,6 +20,8 @@ ripe.Ripe.prototype.init = function(brand, model, options) {
     this.variant = this.options.variant || null;
     this.url = this.options.url || "https://sandbox.platforme.com/api/";
     this.parts = this.options.parts || {};
+    this.initials = "";
+    this.engraving = null;
```
joamag: let's not use the engraving term in the SDK, just initials and profiles

joamag:
@acaldas let's review this manually tomorrow

acaldas:
@joamag the profile is now set using a profileBuilder function. How should the initials be transformed? Like when they have to be uppercase or have dots.

src/python/ripe_demo/static/js/ripe.js
```diff
@@ -1532,6 +1568,21 @@ ripe.Image.prototype.setFrame = function(frame, options) {
     this.update();
 };
 
+ripe.Image.prototype.setInitials = function(initials, options) {
```
joamag: maybe set the engraving together with initials.

PR #41, opened 2018-05-07
added support for optional parts on restrictions

acaldas:
https://github.com/ripe-tech/ripe-core/issues/2967


joamag:
@acaldas need to review this over a call

PR #43, opened 2018-05-08
fixed highlight on wrong parts

acaldas:


src/js/visual/configurator.js
```diff
@@ -46,9 +46,13 @@ ripe.Configurator.prototype.init = function() {
     // highlight operation (considers only the default ones)
     this.partsList = [];
     this.owner.getConfig(function(config) {
-        var defaults = config.defaults;
         this.hiddenParts = config.hidden;
-        this.partsList = Object.keys(defaults);
```
joamag: this was the issue as the order was wrong right ?

acaldas: No, the parts list is in alphabetic order. The problem was that the defaults contains hidden parts like the patch. And on Regent's case it as ```metal_caps```which no longer exists.

joamag: this should definitly not pass the ripe building process, I need to create a verification for this ...

joamag:
Once this build is complete (if it succeeds) re-test this and see if the "fix" is still required https://jenkins-fast.platforme.com/job/3d-build/921/

joamag:
@acaldas please re-test with this build https://jenkins-fast.platforme.com/job/3d-build/922/console

acaldas:
@joamag It's working now

may-10-2018 10-00-02

joamag:
Cool, closing the issue then

PR #45, opened 2018-05-17
Acaldas/browser

acaldas:
Added support for import with ```require``` on the client side.


src/js/base/api.js
```diff
@@ -1,4 +1,4 @@
-if (typeof window === "undefined" && typeof require !== "undefined") {
+if (typeof require !== "undefined") {
```
joamag: why this removal ?

acaldas: With browserify this code will run on the browser, so the window will not be undefined.

joamag: What about the guys that don't use browserify :) ?

joamag: we need a better condition

joamag:
@acaldas any ideas on how to detect browserify ?

acaldas:
@joamag I think the best way is to use a gulp plugin that moves the ripe.js and ripe.css files from the ```node_modules/ripe-sdk``` folder to the ```dist``` folder and include them.

PR #47, opened 2018-05-30
demo - initials container 

hugo-gomes:
hide initials container when no initials defined


joamag:
should use `mode=header,configurator` instead

PR #53, opened 2018-06-28
added support for the undo operation

acaldas:
https://github.com/ripe-tech/ripe-core/issues/3548


src/js/base/ripe.js
```diff
@@ -259,6 +275,15 @@ ripe.Ripe.prototype.update = function(state) {
         );
 };
 
+ripe.Ripe.prototype.undo = function() {
```
joamag: document JSDoc

src/js/base/ripe.js
```diff
@@ -259,6 +275,15 @@ ripe.Ripe.prototype.update = function(state) {
         );
 };
 
+ripe.Ripe.prototype.undo = function() {
+    var parts = this.partsHistory.pop();
+    parts && this.setParts(parts);
+};
+
+ripe.Ripe.prototype.canUndo = function() {
```
joamag: document JSDoc

src/js/base/ripe.js
```diff
@@ -153,22 +157,33 @@ ripe.Ripe.prototype.setOptions = function(options) {
 };
 
 ripe.Ripe.prototype.setPart = function(part, material, color, noUpdate) {
-    this._setPart(part, material, color, true);
+    if (noUpdate !== true) {
```
joamag: why this condition ?

joamag: the `noUpdate` is now a bit outdated in the sense that if one wants to call the no update it's better off calling just the `_set_part` right ?

maybe it's better to just do

```
if(noUpdate) {
    return this._setPart(part, material, color);
}
```

at the beginning of the function

src/js/base/ripe.js
```diff
@@ -35,6 +35,10 @@ ripe.Ripe.prototype.init = function(brand, model, options) {
     var hasParts = this.parts && Object.keys(this.parts).length !== 0;
 
     this.ready = hasParts;
+
+    this.bind("pre_parts", function() {
```
joamag: why not post_parts ?

acaldas: The idea here is that partsHistory would contain all the previous states, so the `pre_parts` event is used to store the current configuration before a change. The `post_parts` event contains the parts after the changes.

joamag: ok so the stack does not contain the current state right ?

acaldas: Right, only the previous states.

joamag: yeahh, and I know that the current implementation. But I think we should improve it and have a sliding strategy (stack pointer) so that we can have undo and redo... That would be cool

PR #64, opened 2018-08-28
only set image dimensions when configured

acaldas:
Currently, if width and height are not set then they are set as 0. This doesn't allow the element to automatically fit the image. This is useful for initials images with viewports.


src/js/ripe.js
```diff
@@ -3418,8 +3418,12 @@ ripe.Image.prototype.update = function(state) {
 
     // updates the image DOM element with the values of the image
     // including requested size and URL
-    this.element.width = width;
-    this.element.height = height;
+    if (width !== null) {
```
joamag: validate with if(width) so that we can handle `0, undefined, null and ""` values :)

PR #82, opened 2019-02-08
Fixed configurator deinit

acaldas:


joamag:
@acaldas how was this discovered? Bug?

PR #85, opened 2019-02-11
improved optional parts handling

acaldas:


src/js/base/ripe.js
```diff
@@ -385,9 +385,21 @@ ripe.Ripe.prototype._getState = function() {
...
 ripe.Ripe.prototype._setPart = function(part, material, color) {
     var value = this.parts[part] || {};
-    value.material = material;
-    value.color = color;
-    this.parts[part] = value;
+
+    // if the material or color are not set then this
+    // is considered a removal operation and the part
+    // is removed from the parts structure
+    if (!material || !color) {
```
joamag: We must validate against mandatory parts, because we should not be able to remove mandatory parts (raise exception)

joamag: Let´s validate this against the currently loaded config

PR #89, opened 2019-02-14
added resolve product id and swatch url builder methods

acaldas:


src/js/api/base.js
```diff
@@ -0,0 +1,36 @@
+if (typeof require !== "undefined") {
+    // eslint-disable-next-line no-redeclare
+    var base = require("../base");
+    // eslint-disable-next-line no-redeclare
+    var ripe = base.ripe;
+}
+
+ripe.Ripe.prototype.resolveProductId = function(productId, options, callback) {
```
joamag: this should be in the config.js file and not here

src/js/api/base.js
```diff
@@ -0,0 +1,36 @@
...
+    options = options || {};
+    const brand = options.brand === undefined ? this.brand : options.brand;
+    const format = options.format === undefined ? "png" : options.format;
+    const query = this._buildQuery({
+        material: material,
+        color: color,
+        brand: brand,
+        format: format
+    });
+    return `${this.url}swatch?${query}`;
```
joamag: is this the standard way of doing this ?

src/js/api/config.js
```diff
@@ -0,0 +1,23 @@
+if (typeof require !== "undefined") {
+    // eslint-disable-next-line no-redeclare
+    var base = require("../base");
+    // eslint-disable-next-line no-redeclare
+    var ripe = base.ripe;
+}
+
+ripe.Ripe.prototype.resolveProductId = function(productId, options, callback) {
```
joamag: name this **configResolve**, we need to start making the API methods respect a standard in the naming to avoid collision of names

src/js/base/api.js
```diff
@@ -325,6 +351,12 @@ ripe.Ripe.prototype._getMaskURL = function(options) {
     return options.url + "?" + this._buildQuery(options.params);
 };
 
+ripe.Ripe.prototype._getSwatchURL = function(material, color, options) {
```
joamag: do we have this parameter standard anywhere else?

acaldas: We have for methods like `signin(username, password, options)` or `getOrder(number, options, callback)`

joamag: But in terms of URL generation is there any that do receive more than the options?

acaldas: No, _getMaskURL and _getImageURL only receive options

src/js/base/api.js
```diff
@@ -325,6 +351,12 @@ ripe.Ripe.prototype._getMaskURL = function(options) {
     return options.url + "?" + this._buildQuery(options.params);
 };
 
+ripe.Ripe.prototype._getSwatchURL = function(material, color, options) {
+    options = Object.assign({ material: material, color: color }, options);
```
joamag: remove this from here and move it maybe to the caller

PR #101, opened 2019-04-22
Fixed sync config check

acaldas:


joamag:
@acaldas maybe it's a good idea to create a unit test for this case

PR #107, opened 2019-06-17
Add API methods for localizing sizes

gcandal:


gcandal:
@joamag while doing this I noticed that `ripe.Ripe.prototype.locale` is defined twice:
- src/js/api/locale.js 
- src/js/base/ripe.js (this one "wins")

Meaning that `ripe.locale("stuff", "en_us")` will result in an error since `locale` will return `en_us` or something like that.

src/js/api/locale.js
```diff
@@ -21,7 +21,7 @@ ripe.Ripe.prototype.locale = function(value, locale, options, callback) {
 /**
  * Localizes a list of values to the provided locale.
  *
- * @param value The values to be localized.
+ * @param values The values to be localized.
  * @param locale The locale to localize the value to.
```
joamag: data type missing in docs

src/js/api/size.js
```diff
@@ -164,3 +164,101 @@ ripe.Ripe.prototype.nativeToSizeB = function(scales, values, genders, options, c
...
+ *
+ * @param {String} scale The scale which one wants to convert from.
+ * @param {Number} value The value which one wants to convert.
+ * @param {String} gender The gender of the scale and value to be converted.
+ * @param {Object} options An object of options to configure the request.
+ * @param {Function} callback Function with the result of the request.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
+ripe.Ripe.prototype.sizeToLocale = function(scale, value, gender, options, callback) {
+    return this.sizeToLocaleB(scale, value, gender, options, callback);
```
joamag: This is not correct b is for bulk

gcandal: True but it can actually handle both. Is that a problem? Anyway, will change to `[value]`, etc.

src/js/api/size.js
```diff
@@ -164,3 +164,101 @@ ripe.Ripe.prototype.nativeToSizeB = function(scales, values, genders, options, c
...
+ *
+ * @param {String} scale The scale which one wants to convert from.
+ * @param {Number} value The value which one wants to convert.
+ * @param {String} gender The gender of the scale and value to be converted.
+ * @param {Object} options An object of options to configure the request.
+ * @param {Function} callback Function with the result of the request.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
+ripe.Ripe.prototype.sizeToLocale = function(scale, value, gender, options, callback) {
+    return this.sizeToLocaleB([scale], [value], [gender], options, callback);
```
joamag: This is an API endpoint please use the correct route

joamag:
@gcandal please create unit tests for these methods

PR #109, opened 2019-06-26
Fix import of logic.js

gcandal:
The import order matters here since `logic.js` is extending the prototype defined in `ripe.js`.


joamag:
This is already fixed

joamag:
https://github.com/ripe-tech/ripe-sdk/commit/ccc31c37e3969e620a8a9c62da7e62e9ff1a8fa3

gcandal:
@joamag isn't the `require` missing in that commit?

joamag:
yup you're right it's safer that way

PR #115, opened 2019-08-30
Fix concurrent drawFrame calls

gcandal:


gcandal:
Fixed by https://github.com/ripe-tech/ripe-sdk/commit/2d62f01353ac4de155140a2f330439b01f7f9a36

joamag:
@gcandal I'm still "seeing" the visual issue

gcandal:
With the latest ripe-sdk? I've never been able to reproduce it after promises were introduced (just retried doing so).

joamag:
Yup this was fixed with the configanimate option. But not by the promises ;)

PR #118, opened 2019-09-17
Allow removing initials using null

gcandal:


src/js/base/ripe.js
```diff
@@ -462,7 +462,7 @@ ripe.Ripe.prototype.setParts = async function(update, events = true, options = {
  * change should be triggered.
  */
 ripe.Ripe.prototype.setInitials = function(initials, engraving, events = true) {
-    if (typeof initials === "object") {
+    if (typeof initials === "object" && initials !== null) {
```
joamag: Why are initials null ?? the "blank" value for them is "", `null` should not be used for initials

gcandal: Not sure that was the intended behavior since `null` in other places means "remove this" (e.g. optionals). Closing this then.

PR #121, opened 2019-10-29
Support for guess

hugo-gomes:


joamag:
@hugo-gomes we need to test this one more time... and create a very simple unit test

PR #122, opened 2019-10-30
Adding _getConfigInfoOptions tests

hugo-gomes:


test/js/api.js
```diff
@@ -320,6 +320,42 @@ describe("RipeAPI", function() {
         });
     });
 
+    describe("#_getConfigInfoOptions", function() {
```
joamag: this should to a config.js test file...

PR #39, opened 2018-05-03
added size api methods

acaldas:
https://github.com/ripe-tech/ripe-core/issues/3300

PR #136, opened 2019-12-17
#39 Justifications API

NFSS10:
Related to issue: [#39](https://github.com/ripe-tech/ripe-pulse/issues/39)
Description: Can now get the "justifications" list. They can optionally be filtered by "context".


src/js/api/justification.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Gets the justifications list. Can be optionally filtered by context.
+ *
+ * @param {Object} options An object of options to configure the query and
+ * its results, such as:
+ * - 'filters[]' - List of filters that the query will use to, operators such as
+ * ('in', 'not_in', 'like', 'likei', 'llike', 'llikei', 'rlike', 'rlikei', 'contains'),
+ * (eg: 'context:eq:42') would filter by the 'context' that equals to '42'.
```
gcandal: Don't use context as an example, it gets a bit confusing given that it is a parameter.

src/js/api/justification.js
```diff
@@ -0,0 +1,55 @@
...
+ * - 'sort' - List of arguments to sort the results by and which direction
+ * to sort them in (eg: 'id:ascending') would sort by the id attribute in ascending order,
+ * while (eg: 'id:descending')] would do it in descending order.
+ * - 'skip' - The number of the first record to retrieve from the results.
+ * - 'limit' - The number of results to retrieve.
+ * @param {Function} callback Function with the result of the request.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
+ripe.Ripe.prototype.getJustifications = function(context = null, options, callback) {
```
gcandal: Having an optional with a default (context) followed by other parameters without it is a bit weird. Is that the way we're doing for other API methods?

gcandal:
Also, the build seems to be failing, so please take a look at it.

gcandal:
@NFSS10 assign me back once the proper fix is pushed

gcandal:
What happened to the `context`?

NFSS10:
> What happened to the `context`?

Added new changed after we cleared the issue.

PR #137, opened 2019-12-17
Create SKU API added

hugo-gomes:


src/js/api/sku.js
```diff
@@ -0,0 +1,98 @@
...
+    const initials = options.initials === undefined ? this.initials : options.initials;
+    const engraving = options.engraving === undefined ? this.engraving : options.engraving;
+    const initialsExtra =
+        options.initials_extra === undefined && options.initialsExtra === undefined
+            ? this.initialsExtra
+            : options.initialsExtra || options.initials_extra;
+
+    const url = this.url + "skus";
+
+    const spec = {
```
joamag: simplify this code, just use a direct Object POST approach

PR #139, opened 2019-12-18
API configSku added

hugo-gomes:


joamag:
@hugo-gomes make some unit tests around this

PR #144, opened 2019-12-20
#39 Cancelling an order with justification

NFSS10:
Related to issue [#39](https://github.com/ripe-tech/ripe-pulse/issues/39).


gcandal:
As discussed, let's avoid the breaking change.

PR #151, opened 2020-02-18
Add react-native compatibility

gcandal:


src/js/base/compat.js
```diff
@@ -51,28 +55,32 @@ ripe.build = function() {
     return ripe.assign.apply(this, _arguments);
 };
 
+const reactNative = typeof navigator !== "undefined" && navigator.product === "ReactNative";
```
joamag: do we need this variable?

gcandal: No, will inline.

PR #156, opened 2020-03-27
Send event on configurator lowlight.

Benedito0:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-whites/issues/183|
| Decisions | Send an event on configurator lowlight|
| Animated GIF |   |
| Dependencies ||
| |  |

### DIFF
--- a/src/js/visual/configurator.js
+++ b/src/js/visual/configurator.js
@@ -587,6 +587,9 @@ ripe.Configurator.prototype.lowlight = function(options) {
     const frontMask = this.element.querySelector(".front-mask");
     frontMask.classList.remove("highlight");
     this.element.classList.remove("highlight");
+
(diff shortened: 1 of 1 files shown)


src/js/visual/configurator.js
```diff
@@ -587,6 +587,7 @@ ripe.Configurator.prototype.lowlight = function(options) {
     const frontMask = this.element.querySelector(".front-mask");
     frontMask.classList.remove("highlight");
     this.element.classList.remove("highlight");
+    this.trigger("lowlight");
```
gcandal: If this event's counterpart is `highlighted_part` than this should be `lowlighted_part`.

joamag:
@Benedito0 Why do we need this event?

PR #157, opened 2020-04-02
#178 Frames and masks times

NFSS10:
| - | - |
| --- | --- |
| Decisions | Calculates the times the masks and images takes to load and sends that information via the bus |
| Animated GIF | - |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -891,10 +891,12 @@ ripe.Ripe.prototype.update = async function(state, options = {}) {
     if (this.ready && this.usePrice) {
         const timestamp = Date.now();
         this._priceTimestamp = timestamp;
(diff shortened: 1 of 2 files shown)


gcandal:
@NFSS10 we shouldn't make changes to the SDK to support the bandwidth plugin. Also, all the changes you did could be made by patching the method from the outside (@BeeMargarida has already done it in a previous tech plugin). The idea is working inside the box (RIPE White) until its very limit, without going around changing dependencies.

Also, avoid opening PRs that are not in the last mile of getting merged. In this specific case, we didn't even discuss the approach, so what's the point of having a PR? The chances of it getting merged, given that it's not something trivial, are pretty low. If it was a small fix or something it would be a different story. 

Closing this and let's talk about the approach to this problem first.

PR #158, opened 2020-04-02
#178 Events

NFSS10:
| - | - |
| --- | --- |
| Decisions | • Added `new_child` to know when a new child was added to ripe <br>• Added `pre_update` to know when the `update` method will start |
| Animated GIF | - |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -879,9 +879,9 @@ ripe.Ripe.prototype.deselectPart = function(part, options = {}) {
  * the update operation is going to be performed.
  */
 ripe.Ripe.prototype.update = async function(state, options = {}) {
-    state = state || this._getState();
-
     this.trigger("pre_update");
+
+    state = state || this._getState();
 
     const promises = [];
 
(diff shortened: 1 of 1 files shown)


src/js/base/ripe.js
```diff
@@ -811,6 +811,7 @@ ripe.Ripe.prototype.bindConfigurator = function(element, options = {}) {
  */
 ripe.Ripe.prototype.bindInteractable = function(element) {
     this.children.push(element);
+    this.trigger("new_child", element);
```
gcandal, line 814: @joamag do you think this makes sense? Maybe this or `interactable_bound`.

We were considering using this in order to be able to bind to the events of children, if they were added in runtime.

src/js/base/ripe.js
```diff
@@ -811,6 +811,7 @@ ripe.Ripe.prototype.bindConfigurator = function(element, options = {}) {
  */
 ripe.Ripe.prototype.bindInteractable = function(element) {
```
gcandal, line 812: @NFSS10 if we trigger on `bind` we must do the reverse on `unbind`.

src/js/base/ripe.js
joamag: let's just call them "bound" and "unbound"

PR #160, opened 2020-04-03
#178: _loadMask and _loadFrame function events

BeeMargarida:
| - | - |
| --- | --- |
| Decisions | Added `pre_frame` and `post_frame`events. |
| Animated GIF | - |

### DIFF
--- a/src/js/visual/configurator.js
+++ b/src/js/visual/configurator.js
@@ -978,19 +978,19 @@ ripe.Configurator.prototype._loadMask = function(maskImage, view, position, opti
         color: backgroundColor
     });
     if (draw && maskImage.dataset.src === url) {
-        this.trigger("post_mask", maskImage, view, position);
+        this.trigger("post_mask", maskImage, view, position, false);
         setTimeout(() => {
             this._drawMask(maskImage);
         }, 150);
     } else {
         maskImage.onload = draw
             ? () => {
-                  this.trigger("post_mask", maskImage, view, position);
+                  this.trigger("post_mask", maskImage, view, position, true);
                   setTimeout(() => {
                       this._drawMask(maskImage);
                   }, 150);
               }
-            : () => this.trigger("post_mask", maskImage, view, position);
+            : () => this.trigger("post_mask", maskImage, view, position, true);
         maskImage.onerror = () => {
             maskImage.removeAttribute("src");
         };
--- a/src/js/visual/configurator.js
+++ b/src/js/visual/configurator.js
@@ -914,7 +914,7 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options 
         if (isReady) {
             await this._drawFrame(image, animate, duration);
         }
-        this.trigger("post_frame", view, position, options, true);
+        this.trigger("post_frame", view, position, options, false);
         return;
     }
 
@@ -952,7 +952,7 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options 
     // we're sure everything is currently loaded
     await imagePromise;
 
-    this.trigger("post_frame", view, position, options, false);
+    this.trigger("post_frame", view, position, options, true);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/js/visual/configurator.js
```diff
@@ -972,17 +978,22 @@ ripe.Configurator.prototype._loadMask = function(maskImage, view, position, opti
...
             ? () => {
+                  this.trigger("post_mask", maskImage, view, position);
                   setTimeout(() => {
                       this._drawMask(maskImage);
                   }, 150);
               }
-            : null;
+            : () => {
+                  this.trigger("post_mask", maskImage, view, position);
+                  return null;
```
gcandal, line 995: No need, you can just turn this into a 1-liner

src/js/visual/configurator.js
```diff
@@ -912,6 +914,7 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options
         if (isReady) {
             await this._drawFrame(image, animate, duration);
         }
+        this.trigger("post_frame", view, position, options);
```
gcandal, line 917: Maybe we should include a boolean in the arguments saying if this was a no-op (the same applies to the mask when it doesn't actually load a new image).

src/js/visual/configurator.js
```diff
@@ -948,12 +951,15 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options
...
     await imagePromise;
+
+    this.trigger("post_frame", view, position, true, options);
 };
 
 /**
  * @ignore
  */
 ripe.Configurator.prototype._loadMask = function(maskImage, view, position, options) {
+    this.trigger("pre_mask", maskImage, view, position);
```
gcandal, line 962: Similarly to `pre_frame` this must also include `options`.

src/js/visual/configurator.js
```diff
@@ -972,17 +978,19 @@ ripe.Configurator.prototype._loadMask = function(maskImage, view, position, opti
         color: backgroundColor
     });
     if (draw && maskImage.dataset.src === url) {
+        this.trigger("post_mask", maskImage, view, position, false);
```
gcandal, line 981: Same as above

src/js/visual/configurator.js
```diff
@@ -912,6 +914,7 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options
         if (isReady) {
             await this._drawFrame(image, animate, duration);
         }
+        this.trigger("post_frame", view, position, false, options);
```
gcandal, line 917: Maybe move the `false` to the end so that we have (event name, operation args..., op/no-op).

gcandal:
With https://github.com/ripe-tech/ripe-sdk/commit/5966310ef17c5eec4c17ce4cb8e5d1ec11ca874d this one is no longer needed, right @NFSS10 ?

BeeMargarida:
> With [5966310](https://github.com/ripe-tech/ripe-sdk/commit/5966310ef17c5eec4c17ce4cb8e5d1ec11ca874d) this one is no longer needed, right @NFSS10 ?
@gcandal We think it is still needed since the `post_udpate` waits for all the updates from its children, given the total time. In this case, we want to catch the time each frame takes to update.

gcandal:
@BeeMargarida ok, please remove the masks part and inline the value of `NO_OP` rather than using a variable then.

src/js/visual/configurator.js
```diff
@@ -912,6 +914,7 @@ ripe.Configurator.prototype._loadFrame = async function(view, position, options
         if (isReady) {
             await this._drawFrame(image, animate, duration);
         }
+        this.trigger("post_frame", view, position, options, true);
```
gcandal, line 917: Aren't these `true`/`false` the opposite of what they are in `post_update`?
- True: updated happened
- False: "cache" hit / no-op

src/python/ripe_demo/static/js/main.js
```diff
@@ -4,7 +4,7 @@
  */
 var FACES = ["side", "top", "front"];
 
-window.onload = function() {
```
gcandal, line 7: I think you need to `npm i`, you aren't using the latest dependencies that's why this happens. In any case, if something like this must be done in the future, please include it in a separate PR.

PR #166, opened 2020-04-30
Fix mask error handler

gcandal:
| - | - |
| --- | --- |
| Decisions | `setAttribute` was supposed to clean the `src` on the mask `` that triggered the error, but was called on `self` which doesn't even define such a method. |


src/js/visual/configurator.js
```diff
@@ -606,7 +606,7 @@ ripe.Configurator.prototype.highlight = function(part, options = {}) {
         self.trigger("highlighted_part", part);
     };
     this.frontMaskError = function() {
-        self.setAttribute("src", "");
+        frontMask.setAttribute("src", "");
```
joamag, line 609: use `this` instead

PR #167, opened 2020-05-04
#341 Fixes hanging white, when removing "Image"

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/341 |
| Dependencies | -- |
| Decisions | • **The symptoms of the problem:**<br>In the same build, when changing models with different number of thumbnails, in this case in brand "dummy", from the model "macho", "cube" or "bill", to the model "dummy", which will go from 3 thumbnails to 2 thumbails, the model can't be switched back. (See  [#341](https://github.com/ripe-tech/ripe-white/issues/341) for more details)<br>• **What the root cause issue was:**<br>`cancel` operation in two of the `Image`s would not be resolved so they would hang ripe-white. It would hang here:https://github.com/ripe-tech/ripe-sdk/blob/3450aa661f7fee03a9003c29ac89b64ffc18dc92/src/js/visual/image.js#L153 which would then hang here: https://github.com/ripe-tech/ripe-sdk/blob/3450aa661f7fee03a9003c29ac89b64ffc18dc92/src/js/base/ripe.js#L1130 and the flow would not go further from here: https://github.com/ripe-tech/ripe-sdk/blob/3450aa661f7fee03a9003c29ac89b64ffc18dc92/src/js/base/ripe.js#L1195 and here:https://github.com/ripe-tech/ripe-sdk/blob/3450aa661f7fee03a9003c29ac89b64ffc18dc92/src/js/base/ripe.js#L311<br>• Why this change fixes the root cause:<br>When destroying the thumbnails, `Image` `deinit` is called: https://github.com/ripe-tech/ripe-sdk/blob/3450aa661f7fee03a9003c29ac89b64ffc18dc92/src/js/visual/image.js#L198 As the callbacks are removed before being resolved, the promises would hang in limbo. Resolving these pending promises, fixes this bug. |
| Animated GIF | White-Dummy_dummy_fix |

### DIFF
--- a/src/js/visual/image.js
+++ b/src/js/visual/image.js
@@ -195,7 +195,8 @@ ripe.Image.prototype.cancel = async function(options = {}) {
  * it should stop responding to updates so that any necessary
  * cleanup operations can be executed.
  */
-ripe.Image.prototype.deinit = function() {
+ripe.Image.prototype.deinit = async function() {
+    await this.cancel();
     this._unregisterHandlers();
     this._observer = null;
     this.initialsBuilder = null;
@@ -269,7 +270,6 @@ ripe.Image.prototype._registerHandlers = function() {
 ripe.Image.prototype._unregisterHandlers = function() {
     if (this.loadListener) this.element.removeEventListener("load", this.loadListener);
     if (this.errorListener) this.element.removeEventListener("error", this.errorListener);
-    if (this._loadedCallback) this._loadedCallback({ canceled: true });
     if (this.loadedHandler) this.unbind("loaded", this.loadedHandler);
     if (this.errorHandler) this.unbind("error", this.errorHandler);
     if (this._observer) this._observer.disconnect();
--- a/src/js/visual/image.js
+++ b/src/js/visual/image.js
@@ -269,6 +269,7 @@ ripe.Image.prototype._registerHandlers = function() {
 ripe.Image.prototype._unregisterHandlers = function() {
     if (this.loadListener) this.element.removeEventListener("load", this.loadListener);
     if (this.errorListener) this.element.removeEventListener("error", this.errorListener);
+    if (this._loadedCallback) this._loadedCallback({ canceled: true });
     if (this.loadedHandler) this.unbind("loaded", this.loadedHandler);
     if (this.errorHandler) this.unbind("error", this.errorHandler);
     if (this._observer) this._observer.disconnect();
--- a/src/js/base/interactable.js
+++ b/src/js/base/interactable.js
@@ -57,7 +57,7 @@ ripe.Interactable.prototype.update = async function(state, options = {}) {};
  * it should stop responding to updates so that any necessary
  * cleanup operations can be executed.
  */
-ripe.Interactable.prototype.deinit = function() {
+ripe.Interactable.prototype.deinit = async function() {
     this.owner = null;
 
     ripe.Observable.prototype.deinit.call(this);
--- a/src/js/base/observable.js
+++ b/src/js/base/observable.js
@@ -102,7 +102,7 @@ ripe.Observable.prototype.runCallbackNoWait = async function(event, ...args) {
  * The deinitializer of the class, called whenever this
  * observable ceases its activity.
  */
-ripe.Observable.prototype.deinit = function() {
+ripe.Observable.prototype.deinit = async function() {
     this.callbacks = null;
 };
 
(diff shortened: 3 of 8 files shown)


src/js/visual/image.js
```diff
@@ -269,6 +269,7 @@ ripe.Image.prototype._registerHandlers = function() {
 ripe.Image.prototype._unregisterHandlers = function() {
     if (this.loadListener) this.element.removeEventListener("load", this.loadListener);
     if (this.errorListener) this.element.removeEventListener("error", this.errorListener);
+    if (this._loadedCallback) this._loadedCallback({ canceled: true });
```
NFSS10, line 272: @gcandal This is the fix, the remaining changes were Pretty's fault

gcandal:
@NFSS10 enhance the `Decisions` section with the full explanation of what's going, including at least:
1. The symptoms of the problem
2. What the root cause issue was
3. Why this change fixes the root cause

If necessary, include pointers (links) to relevant parts of the code.

It's essential for non-trivial PRs to contain the full "story" of what happened, both to make the work of reviewers easier (if I was to review this right now I would almost have to repeat the investigation of what you already did) and to serve as an immutable record, in case someone later wants to git blame and understand why this new line exists and what does it do.

gcandal:
@NFSS10 also: please don't forget adding me as a reviewer and using the correct label (bug/enhancement), it's the kind of thing that helps us.

NFSS10:
> @NFSS10 enhance the `Decisions` section with the full explanation of what's going, including at least:
> 
>     1. The symptoms of the problem
> 
>     2. What the root cause issue was
> 
>     3. Why this change fixes the root cause
> 
> 
> If necessary, include pointers (links) to relevant parts of the code.
> 
> It's essential for non-trivial PRs to contain the full "story" of what happened, both to make the work of reviewers easier (if I was to review this right now I would almost have to repeat the investigation of what you already did) and to serve as an immutable record, in case someone later wants to git blame and understand why this new line exists and what does it do.

@gcandal Added more details

src/js/visual/image.js
gcandal: @NFSS10 please revert them. I don't know what's going on, but clearly there shouldn't be a space between `function` and `()`.

gcandal, line 272: @NFSS10 this should probably be `this.cancel()`, no? Right now it's probably the same thing, but in the future it might contain more stuff.

Besides, I'd place that call under `deinit` instead. Right now it's under `unregisterHandlers` which is about unregistering event listeners, so it seems out of place.

After this please review `configurator` for the same class of problems, if I had to guess the same should be happening there, it's just not a problem because people are never calling `deinit` on them.

gcandal: The hard work is done, we're in the last mile 💪

src/js/visual/image.js
```diff
@@ -195,7 +195,8 @@ ripe.Image.prototype.cancel = async function(options = {}) {
  * it should stop responding to updates so that any necessary
  * cleanup operations can be executed.
  */
-ripe.Image.prototype.deinit = function() {
+ripe.Image.prototype.deinit = async function() {
```
gcandal, line 198: @NFSS10 this is not enough, it has to be replicated to all other `deinit` and the functions that call them.

gcandal: @NFSS10 still unresolved.

gcandal:
@NFSS10 take a look at the changes I did:
- Adding `async` to every `deinit`. It's important that all methods that share a common interface really have the same interface. If this wasn't the case this could happen: I'm writing code to handle `Observable#deinit`, so I'll go check on that function's signature and verify that's not `async` therefore I won't `await`. This would be wrong because one of those observables could be an `Image`,  which has an async deinit
- The configurator had the same problem as image, so `cancel` should be there too

gcandal:
@NFSS10 we'll also probably need to add `await` to all our usages of these methods across all repositories, please check that.

joamag:
@gcandal there are tests failing...

gcandal:
@NFSS10 please check the tests, there must be `await`s missing.

NFSS10:
> @NFSS10 please check the tests, there must be `await`s missing.
Should be good now !
imagem

joamag:
We have the first reward winner 🏆 @NFSS10 congrats!

PR #174, opened 2020-06-04
Fixed "message" event

NFSS10:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | `result.messages` method of getting the pair `name`, `value` was being treated similarly as it would an object with `object.entries()`, but it as it is an array it will give values that aren't what we want (see image below), so I fixed the for loop  |
| Image | imagem |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -1638,7 +1638,11 @@ ripe.Ripe.prototype._handleCtx = function(result) {
     if (result.choices && !ripe.equal(result.choices, this.choices)) {
         this.setChoices(result.choices);
     }
-    for (const [name, value] of result.messages) {
+
+    for (let i = 0; i < result.messages.length; i += 2) {
+        const name = result.messages[i];
+        const value = result.messages[i + 1];
+
         this.trigger("message", name, value);
     }
 };


NFSS10:
@gcandal I don't know if `result.messages` should be an `array` or an `object`, either way this fixes for what is implemented in the moment

gcandal:
Nice. Can you please link in Decisions the place in RIPE Core where it is ensured that the messages structure is the one you're adapting to?

gcandal:
@hugo-gomes @joamag are the messages in https://github.com/ripe-tech/builds-static/search?q=messages&unscoped_q=messages well-defined? Is it supposed to be in the `[message1 name, message 1 value, message 2 name, message 2 value]` format?

NFSS10:
@gcandal  Updated decisions with more information

joamag:
@gcandal I'm not sure about those specific cases, but messages are a sequence of message tuples containing the context of the message and the message itself. When in doubt check the developer portal here https://dev.platforme.com/p/specs/build.md#base.py

Capture2

NFSS10:
@gcandal  The build is wrong

https://github.com/ripe-tech/builds-static/blob/1bb818fa9d2d70a34c492c8918bc6d1204413e06/config.static/dummy/static/logic/cube.py#L31

PR #175, opened 2020-06-04
#426 Show on_X messages for 3DB logic.py

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/426 |
| Dependencies | - |
| Decisions | Add global message event |
| Animated GIF | - |

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,7 +5,7 @@
             "name": "Python",
             "type": "python",
             "request": "launch",
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/src/python/ripe_demo/main.py",
             "debugOptions": [
                 "RedirectOutput"
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,7 +5,7 @@
             "name": "Python",
             "type": "python",
             "request": "launch",
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/src/python/ripe_demo/main.py",
             "debugOptions": [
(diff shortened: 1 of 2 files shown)


gcandal:
Why not just use the existing event?

3rdvision:
> Why not just use the existing event?

We want the whole messages object, this way we can display all the message tuples and not just the last one.
This way we can display the messages like:
image

gcandal:
> > Why not just use the existing event?
> 
> We want the whole messages object, this way we can display all the message tuples and not just the last one.
> This way we can display the messages like:
> image

Ok but you have the events for every message, so you just need to catch all `message` events and append them, no?

NFSS10:
> > > Why not just use the existing event?
> > 
> > 
> > We want the whole messages object, this way we can display all the message tuples and not just the last one.
> > This way we can display the messages like:
> > image
> 
> Ok but you have the events for every message, so you just need to catch all `message` events and append them, no?

But we don't want to do that right? We want the new set of messages to overwrite the previous ones.

gcandal:
I don't think there's need for an extra event, you can simply use the existing events: config, part, initials and initials_extra.

Have in mind that adding public API (new methods, arguments, events, etc) must always be done with extreme caution because it's something we'll have to support forever.

NFSS10:
> 
> 
> I don't think there's need for an extra event, you can simply use the existing events: config, part, initials and initials_extra. Any reason why not?


Yes we can, the ideia here was to avoid doing another bind in the future if we would add another feature similar to config, initials, etc


> 
> Have in mind that adding public API (new methods, arguments, events, etc) must always be done with extreme caution because it's something we'll have to support forever.

That makes sense, we will keep that in mind for the future. I will close this PR and adapt https://github.com/ripe-tech/ripe-commons-pluginus/pull/118#event-3412329171

Issue #176, opened 2020-06-15
Unit tests for subscription endpoints


gcandal:
## Description

Create the unit tests for https://github.com/ripe-tech/ripe-sdk/pull/172 , which should be similar to the existing API tests that use a real RIPE Core instance to perform API calls.

PR #172, opened 2020-05-22
Push notifications

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-robin-revamp/issues/11 |
| Dependencies | https://github.com/ripe-tech/ripe-core/pull/4455 |
| Decisions | - |
| Animated GIF | - |

PR #177, opened 2020-06-16
#176 Unit tests for subscription endpoints

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/176 |
| Decisions | Add unit tests to subscription endpoints created for in https://github.com/ripe-tech/ripe-sdk/pull/172 |

### DIFF
--- a/test/js/api/notify-info.js
+++ b/test/js/api/notify-info.js
@@ -6,6 +6,7 @@ const ripe = require("../../../src/js");
 describe("NotifyInfoAPI", function() {
     this.timeout(config.TEST_TIMEOUT);
 
+
     describe("#createDeviceId", function() {
         beforeEach(function() {
             if (!config.TEST_USERNAME || !config.TEST_PASSWORD) {
--- a/test/js/api/notify-info.js
+++ b/test/js/api/notify-info.js
@@ -1,0 +1,60 @@
+const assert = require("assert");
+const uuid = require("uuid");
+const config = require("../config");
+const ripe = require("../../../src/js");
+
+describe("NotifyInfoAPI", function() {
+    this.timeout(config.TEST_TIMEOUT);
+
+    describe("#createDeviceId", function() {
+        beforeEach(function() {
+            if (!config.TEST_USERNAME || !config.TEST_PASSWORD) {
+                this.skip();
+            }
+        });
+
+        it("should be able to create a deviceID", async () => {
+            let result = null;
+
+            const remote = ripe.RipeAPI();
+
+            const deviceId = uuid.v4();
+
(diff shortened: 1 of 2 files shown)


test/js/api/notify-info.js
```diff
@@ -0,0 +1,60 @@
...
+    this.timeout(config.TEST_TIMEOUT);
+
+    describe("#createDeviceId", function() {
+        beforeEach(function() {
+            if (!config.TEST_USERNAME || !config.TEST_PASSWORD) {
+                this.skip();
+            }
+        });
+
+        it("should be able to create a deviceID", async () => {
```
gcandal, line 16: Let's keep it as english, not code 😅 : `device ID`

gcandal, line 16: Test also what happens when you create the same device ID twice. Unit tests are the time to get creative :)

test/js/api/notify-info.js
```diff
@@ -0,0 +1,60 @@
...
+            assert.notStrictEqual(typeof result.sid, undefined);
+
+            result = await remote.createDeviceIdP(deviceId);
+
+            assert.notStrictEqual(typeof result.created, undefined);
+            assert.notStrictEqual(typeof result.email, undefined);
+        });
+    });
+
+    describe("#removeDeviceId", function() {
```
gcandal, line 35: Add test for removing a device ID that doesn't exist.

test/js/api/notify-info.js
gcandal: Also, try adding 2 different device IDs for the same user.

test/js/api/order.js
```diff
@@ -270,4 +270,72 @@ describe("OrderAPI", function() {
...
+            let result = null;
+
+            const remote = ripe.RipeAPI();
+
+            result = await remote.authAdminP(config.TEST_USERNAME, config.TEST_PASSWORD);
+
+            assert.strictEqual(result.username, config.TEST_USERNAME);
+            assert.notStrictEqual(typeof result.sid, undefined);
+
+            result = await remote.subscribeOrderP(4488);
```
gcandal, line 313: Do the tests reset the data between runs? What happens if I run this test and then run `#getOrderSubscription ` with the same user? Won't I be subscribed and make the other test fail?

test/js/api/order.js
```diff
@@ -270,4 +270,72 @@ describe("OrderAPI", function() {
             });
         });
     });
+
```
gcandal, line 273: Try to think of some more scenarios for this, such as subscribing and then unsubscribing, etc.

PR #181, opened 2020-06-29
#504 Add get locale model keys

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-building/issues/504 |
| Decisions | Add method to retrieve locale model keys |

### DIFF
--- a/src/js/api/build.js
+++ b/src/js/api/build.js
@@ -520,6 +520,19 @@ ripe.Ripe.prototype.getLocaleModelP = function(options) {
     });
 };
 
+/**
+ * Retrieves the default locale keys of a specific brand and model
+ * If no brand is defined it retrieves the keys of the owner's current brand.
+ * If no model is defined it retrieves the keys of the owner's current model.
+ * If no version is defined it retrieves the keys of the owner's current version.
+ *
+ * @param {Object} options An object of options to configure the request, such as:
+ * - 'brand' - The brand of the model.
+ * - 'model' - The name of the model.
+ * - 'version' - The build version.
+ * @param {Function} callback Function with the result of the request.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
 ripe.Ripe.prototype.getLocaleModelKeys = function(options, callback) {
     callback = typeof options === "function" ? options : callback;
     options = typeof options === "function" || options === undefined ? {} : options;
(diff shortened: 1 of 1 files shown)


src/js/api/build.js
```diff
@@ -520,6 +520,63 @@ ripe.Ripe.prototype.getLocaleModelP = function(options) {
...
+    const model = options.model === undefined ? this.model : options.model;
+    const version = options.version === undefined ? this.version : options.version;
+    const url = `${this.url}builds/${brand}/locale/keys`;
+    const params = {};
+    if (model !== undefined && model !== null) {
+        params.model = model;
+    }
+    if (version !== undefined && version !== null) {
+        params.version = version;
+    }
```
gcandal, line 549: Can't you reuse `_getLocaleModelOptions`? It gives you some stuff for free, like using the currently configured model, instead of requiring a parameter.

3rdvision: Both `brand`, `model` and `version` are optional parameters as `this.brand`, `this.model` and `this.version` will use currently configured options if none are given. This has also already been tested and getLocaleModelP() is getting the right currently configured brand, model and version without having to pass them through `options` argument.

gcandal: You're right, I hadn't seen your `this.model`

src/js/api/build.js
```diff
@@ -520,6 +520,63 @@ ripe.Ripe.prototype.getLocaleModelP = function(options) {
     });
 };
 
+/**
+ * Retrieves the default locale keys of a specific brand and model
+ * If no brand is defined it retrieves the keys of the owner's current brand.
+ * If no model is defined it retrieves the keys of the owner's current model.
+ * If no version is defined it retrieves the keys of the owner's current version.
```
gcandal, line 527: Move this explanation to where the fields are described below.

src/js/api/build.js
```diff
@@ -520,6 +520,63 @@ ripe.Ripe.prototype.getLocaleModelP = function(options) {
     });
 };
 
+/**
+ * Retrieves the default locale keys of a specific brand and model
```
gcandal, line 524: End with `.`.

src/js/api/build.js
```diff
@@ -520,6 +520,57 @@ ripe.Ripe.prototype.getLocaleModelP = function(options) {
...
+ */
+ripe.Ripe.prototype.getLocaleModelKeys = function(options, callback) {
+    callback = typeof options === "function" ? options : callback;
+    options = typeof options === "function" || options === undefined ? {} : options;
+    const brand = options.brand === undefined ? this.brand : options.brand;
+    const model = options.model === undefined ? this.model : options.model;
+    const version = options.version === undefined ? this.version : options.version;
+    const url = `${this.url}builds/${brand}/locale/keys`;
+    const params = {};
+    if (model !== undefined && model !== null) {
```
joamag, line 541: Create `getLocaleModelKeysOptions()`, just for consistency

PR #186, opened 2020-07-23
Support for invalid frame error handling

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Related to https://github.com/ripe-tech/ripe-sdk-components-vue/issues/1. It is necessary to make the `configurator` solution more resilient to wrong input. |
| Dependencies | -- |
| Decisions | When an invalid frame is given, and exception is transported to the higher level. However, if a valid frame is later given, it will be ignored since the class `animating` is still applied to the `configurator`. It is necessary to catch the exception thrown by `update` when an invalid frame is given and remove that class, not allowing the recursive action to continue and making it possible to later give a valid frame. |
| Animated GIF | **Before:** <br> Jul-23-2020 11-23-17<br> **After:** <br>solution|

### DIFF
--- a/src/js/visual/configurator.js
+++ b/src/js/visual/configurator.js
@@ -357,7 +357,8 @@ ripe.Configurator.prototype.changeFrame = async function(frame, options = {}) {
     }
 
     // in case the safe mode is enabled and there's an animation running
-    // then this request is going to be ignored
+    // then this request is going to be ignored (not possible to change
+    // frame when another animation is running)
     if (safe && this.element.classList.contains("animating")) {
         return;
     }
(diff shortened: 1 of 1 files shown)


src/js/visual/configurator.js
```diff
@@ -527,13 +527,18 @@ ripe.Configurator.prototype.changeFrame = async function(frame, options = {}) {
...
-    );
+    try {
+        await this.update(
+            {},
+            {
+                animate: animate,
+                duration: animate ? duration : 0
+            }
+        );
+    } catch (error) {
```
joamag, line 538: use try/finally instead of re-throw for situations where cleanup is required (in this case the removal of the no-drag and animating classes)

PR #193, opened 2020-08-05
Fix crop update

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | |
| Decisions | Since `crop` is a boolean, it is necessary to make a different update logic. |
| Animated GIF | |


gcandal:
Maybe this should be applied to numbers too, what do you think?

BeeMargarida:
> Maybe this should be applied to numbers too, what do you think?

@gcandal It doesn't have impact in the logic, but the code might become more uniform and understandable.

joamag:
@BeeMargarida let's backport this to the `Configurator`

PR #209, opened 2020-11-09
Add getMesh

gcandal:
Group:
- https://github.com/ripe-tech/ripe-sdk/pull/209
- https://github.com/ripe-tech/ripe-core/pull/4494
- https://github.com/ripe-tech/ripe-compose/pull/99

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -621,4 +621,4 @@ ripe.Ripe.prototype._getMeshOptions = function(options = {}) {
         method: "GET",
         params: params
     });
-}
+};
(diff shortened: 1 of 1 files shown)


joamag (approved):

joamag:
@gcandal We'll have to have a `getMeshUrl` to be used when the loading of the mesh is not to be controlled directly by "us"

gcandal:
> @gcandal We'll have to have something like `getMeshUrl` to be used when the loading of the mesh is not to be controlled directly by "us"

@joamag is the idea having something like below?

```json
{
    "meshes": {
        "$base": {
            "url": "https://...",
            "polycount": "high"
        }
    }
}
```

That would imply having the same signature for arguments in `getMeshUrl` and `getMesh`. Is that what you had in mind?

joamag:
No, the idea is on the client-side have a way of just getting the URL to obtain the mesh and not the mesh payload directly

PR #214, opened 2020-11-12
Use node_modules dependencies

gcandal:
| - | - |
| --- | --- |
| Issue | Dependencies were being fetched from CDNs rather than included in the bundle. |
| Decisions | - Includes dependencies needed for CSR all the time, even if for some use cases that's irrelevant and only increases the bundle size. Eventually, we could have several bundles that include different capabilities (sort of like pconvert-rust with the Python and WASM bindings).<br/>- Currently is not using minimized dependencies, but that's easy to include as soon as we validate this approach. <br/> - Didn't know from which branch to start from so used `pr/csr-initials`. <br/> - Bundle before: 881K. Bundle after: 3.2M. Again, not minimized. |


joamag:
@gcandal damm we went from 387kb in master to 1.8MB with all of this :(. I believe we'll have to deploy different versions of ripe some enabling CSR and some not allowing it

joamag:
with gzip we're talking about 400KB, so it may work

joamag:
@gcandal let's have a quick chat of 10 min to discuss this one, as this is major! 💪

gulpfile.js
```diff
@@ -11,20 +11,32 @@ const eslint = require("gulp-eslint");
...
 const sourcemaps = require("gulp-sourcemaps");
+const mergeStream = require("merge-stream");
 const _package = require("./package.json");
 
 const paths = {
     mainjs: "dist/ripe.js",
     mainmap: "dist/ripe.js.map",
     maincss: "src/css/ripe.css",
-    scripts: "src/js/**/*.js",
+    scripts: ["src/js/**/*.js"],
```
joamag: no like

joamag (approved):

PR #213, opened 2020-11-11
#172 Support for more image options

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-design/issues/172 |
| Dependencies | -- |
| Decisions | Support more image options. (Saw which ones were missing from here https://github.com/ripe-tech/ripe-compose/blob/master/src/ripe_compose/controllers/web/composition.py#L150-L210) |
| Animated GIF | -- |

PR #215, opened 2020-11-13
#172 Image options test

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/pull/213#issuecomment-726264057 |
| Dependencies | -- |
| Decisions | • Added `should generate the correct image URL` test <br>• Fixed bad URL when passing `offsets` and `curve` |
| Screenshot | imagem |

### DIFF
--- a/src/js/base/api.js
+++ b/src/js/base/api.js
@@ -812,11 +812,11 @@ ripe.Ripe.prototype._getImageOptions = function(options = {}) {
     }
 
     if (options.offsets !== undefined && options.offsets !== null) {
-        params.offsets = options.offsets;
+        params.offsets = JSON.stringify(options.offsets);
     }
 
     if (options.curve !== undefined && options.curve !== null) {
-        params.curve = options.curve;
+        params.curve = JSON.stringify(options.curve);
     }
 
     const url = `${this.url}compose`;
--- a/test/js/visual.js
+++ b/test/js/visual.js
@@ -49,13 +49,12 @@ describe("Visual", function() {
             assert.strictEqual(imageElement.src.includes("frame=side-10"), true);
         });
 
-        it("should generate the correct image URL", () => {
+        it("should generate the correct image URL", async () => {
             const dom = new jsdom.JSDOM("<img id='image' />");
             const imageElement = dom.window.document.getElementById("image");
             const instance = new base.ripe.Ripe("myswear", "vyner");
+            await instance.isReady();
             const image = instance.bindImage(imageElement, {
-                brand: "myswear",
-                model: "vyner",
                 rotation: 20,
                 flip: true,
                 mirror: true,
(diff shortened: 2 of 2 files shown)


src/js/base/api.js
```diff
@@ -812,11 +812,11 @@ ripe.Ripe.prototype._getImageOptions = function(options = {}) {
     }
 
     if (options.offsets !== undefined && options.offsets !== null) {
-        params.offsets = options.offsets;
+        params.offsets = JSON.stringify(options.offsets);
     }
 
     if (options.curve !== undefined && options.curve !== null) {
-        params.curve = options.curve;
+        params.curve = JSON.stringify(options.curve);
```
gcandal, line 819: How did you test this?

These example URLs are not using JSON for that field:
- https://ripe-core-sbx.platforme.com/test/config/render?brand=kirkwood&model=beya_loafer&initials=AAA&initials_debug=1&curve=[[0.2,%200.2],%20[0.7,%200.2],%20[0.2,%200.5],%20[0.7,%200.5]]&cache=0&crop=1
- https://ripe-core-sbx.platforme.com/compose?brand=kirkwood&model=beya_loafer&initials=AAA&initials_debug=1&curve=[[0.2,%200.2],%20[0.7,%200.2],%20[0.2,%200.5],%20[0.7,%200.5]]&cache=0&crop=1

NFSS10: The generated URL wasn't working, I needed to do this so that the URL is generated correctly

gcandal: Can you show me an example of a compose URL working using JSON-encoding for `curve`? Does it look similar to what I posted?

gcandal (approved):

test/js/visual.js
```diff
@@ -49,6 +49,62 @@ describe("Visual", function() {
...
+                    [0.2, 0.5],
+                    [0.7, 0.5]
+                ]
+            });
+            assert.strictEqual(image._url, null);
+
+            image.update();
+            assert.strictEqual(
+                image._url,
+                "https://sandbox.platforme.com/api/compose?algorithm=mask_top&background=ff00ff&bounding_box=2000&bounding_box=2000&curve=%5B%5B0.2%2C0.2%5D%2C%5B0.7%2C0.2%5D%2C%5B0.2%2C0.5%5D%2C%5B0.7%2C0.5%5D%5D&debug=true&engine=base&flip=true&font_size=50&font_spacing=10&font_trim=true&font_weight=900&initials_algorithm=multiplicative&initials_align=left&initials_blend_color=222222&initials_color=ff0000&initials_height=20&initials_image_flip=true&initials_image_rotation=30&initials_opacity=0.7&initials_rotation=20&initials_vertical=top&initials_viewport=1&initials_viewport=1&initials_viewport=20&initials_viewport=20&initials_width=20&initials_x=2&initials_y=2&initials_z_index=5&line_height=30&mirror=true&offsets=%7B%220%22%3A%5B0%2C10%5D%2C%221%22%3A%5B-1%2C-10%5D%7D&rotation=20&shadow=true&shadow_color=0000ff&shadow_offset=20"
```
gcandal, line 104: Shouldn't the model and brand be here? This URL doesn't work:

image

gcandal (changes requested):

src/js/base/api.js
gcandal: Ok apparently it's correct, just checked

test/js/visual.js
NFSS10: It needs an brand and model and it works.
I think its intended for `_url` to not have as the test is made, so I think the test is correct.

If you check https://github.com/ripe-tech/ripe-sdk/pull/213 no changes related to brand and model were made.

gcandal: How can the test be correct if the URL you're checking produces a server error? I'm not saying you introduced any regression, but clearly there is something fishy here that needs fixing otherwise we have a test which isn't testing what it is supposed to.

NFSS10: I will tweak it then

gcandal (approved):

joamag (approved):

PR #218, opened 2020-11-16
#677: on_part not respected

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/677 |
| Dependencies | -- |
| Decisions | **Original problem**: `_handleCtx` function called after [`onPartP`](https://github.com/ripe-tech/ripe-sdk/blob/337af50f0ca3402ecbd40669d6c7a04a87fcf4be/src/js/base/ripe.js#L167) was not handling the removal of parts (added that). However, this brought a **new problem**, which happens when going from `no legs` (no shadow) to `legs` (with shadow). The cycle is: <br><br>click picker -> `setPart` -> `_setPart` -> trigger `"part"` -> **restriction** -> `_solveRestrictions` -> [`customization.length == 0` but `isComplete` returns `false`](https://github.com/ripe-tech/ripe-sdk/blob/f06977fc1e1857c6ce10f7d11890872460954bef/src/js/plugins/restrictions.js#L152) (`shadow` was still not added to parts) -> **error** because it keeps trying to check the customization but there are no more parts. <br><br> With this, I changed the check to only `customization.length === 0`, which is a bit risky. <br> The shadow is then added in the `_handleCtx` function, which runs after the restrictions. <br> The `_setPart` method only hads the legs and has no way of knowing that it should also add the shadow. I don't see another workaround except this addition only being done in the [`_handleCtx`](https://github.com/ripe-tech/ripe-sdk/blob/0d17c73741e1acb27a0d557d7116064298e5ab33/src/js/base/ripe.js#L1740), but it would be possible to maybe fix this if the [`onPartP`](https://github.com/ripe-tech/ripe-sdk/blob/337af50f0ca3402ecbd40669d6c7a04a87fcf4be/src/js/base/ripe.js#L167) logic ran first and then the restrictions (but it does not make sense logic wise).  |
| Animated GIF |  Peek 2020-11-16 17-55|


gcandal:
Closing as we'll be using a slightly different approach. I'll send you the commit later so you can take a look at the difference, but it's essentially what we discussed

joamag (approved):

PR #221, opened 2020-11-26
#39: Added currency to structure

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Derived from https://github.com/ripe-tech/ripe-sdk-components-vue/pull/43#issuecomment-734344558 |
| Dependencies | -- |
| Decisions | Added currency to `structure`. <br> Switched `isEmpty` to keys.length > 0, because it was causing error.  |
| Animated GIF | -- |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -554,6 +554,7 @@ ripe.Ripe.prototype.getStructure = async function(safe = true) {
     if (this.parts && Object.keys(this.parts).length > 0) {
         structure.parts = this.parts;
     }
+    if (this.currency) structure.currency = this.currency;
     if (this.initials) structure.initials = this.initials;
     if (this.engraving) structure.engraving = this.engraving;
     if (this.initialsExtra && Object.keys(this.initialsExtra).length > 0) {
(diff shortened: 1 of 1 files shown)


gcandal:
@BeeMargarida this is probably not the way to go since we're messing with the meaning of `structure`, which should only represent a customization+personalization and not the full set of SDK settings.

What I was thinking was changing `setStructure` so that it would receive an extra `options` argument as well, similarly to what already happens with `fromStructure`.

Regarding the `isEmpty` part, can you detail what was the problem?

BeeMargarida:
> @BeeMargarida this is probably not the way to go since we're messing with the meaning of `structure`, which should only represent a customization+personalization and not the full set of SDK settings.
> 
> What I was thinking was changing `setStructure` so that it would receive an extra `options` argument as well, similarly to what already happens with `fromStructure`.
> 
> Regarding the `isEmpty` part, can you detail what was the problem?

Oh, got it, I misunderstood. Will correct!
Regarding the isEmpty, it was throwing the error that it was not a function. I don't think we are using lodash, and didn't see any other uses, in other places it is using the `Object.keys(..).length > 0`.

gcandal:
You're right, `isEmpty` is for strings. Can you open that as a separate PR? It's probably an insta-merge.

src/js/base/ripe.js
```diff
@@ -56,7 +56,7 @@ ripe.RipeBase = function(brand, model, options = {}) {
  */
 ripe.Ripe.fromStructure = async function(structure, safe = true, options = {}) {
     const instance = new ripe.Ripe(options);
-    await instance.setStructure(structure, safe);
+    await instance.setStructure(structure, safe, options);
```
gcandal, line 59: I don't think you need this here, given that in the line above you're already passing the options

gcandal (changes requested):

gcandal:
Let's do this differently

PR #224, opened 2020-12-03
Brand getLogic endpoints

BeeMargarida:
| - | - |
| --- | --- |
| Dependencies | -- |
| Decisions | • Function to connect to the endpoint that returns the logic `js` script for the build.  |
| Animated GIF | -- |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -665,9 +665,7 @@ ripe.Ripe.prototype._getLogicOptions = function(options = {}) {
         params.method = method;
     }
     if (args !== undefined && args !== null) {
-        for (const [key, value] of Object.entries(args)) {
-            params[key] = value;
-        }
+        Object.assign(params, args);
     }
     return Object.assign(options, {
         url: url,
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -681,7 +681,7 @@ ripe.Ripe.prototype._getLogicOptions = function(options = {}) {
         params.version = version;
     }
     if (format !== undefined && format !== null) {
-        params.format = "js";
+        params.format = format;
     }
     if (method !== undefined && method !== null) {
         params.method = method;
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -339,6 +339,8 @@ ripe.Ripe.prototype.runLogicP = function(options) {
  *  - 'model' - The name of the model
  *  - 'version' - The version of the build, defaults to latest
  *  - 'format' - The format of the logic script ("js" or "py")
+ *  - 'method' - The method of the logic module of the model
+ *  - 'args' - The arguments to pass to the method, as an object
  * @param {Function} callback Function with the result of the request.
  * @returns {XMLHttpRequest} The logic script of the provided model.
  */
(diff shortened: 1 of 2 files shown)


src/js/api/brand.js
```diff
@@ -624,6 +661,29 @@ ripe.Ripe.prototype._runLogicOptions = function(options = {}) {
     });
 };
 
+/**
+ * @ignore
+ */
+ripe.Ripe.prototype._getLogicOptions = function(options = {}) {
```
gcandal, line 667: Isn't `method` missing? See: https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/controllers/api/brand.py#L227

And since we're working on this, is https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/controllers/api/brand.py#L243 already in RIPE SDK?

BeeMargarida: I din't implement the method name because `runLogic` (which is the `logic_method` endpoint) already does that. The getLogic returns the full script.

gcandal: Sure but if it exists in the RIPE Core it must exist here too, there's no reason for the API client to restrict what can be done with the API it represents.

gcandal: Regarding the second point I can see the answer is `yes`

BeeMargarida: @gcandal While I was implementing this I noticed a problem in the logic endpoint. It is impossible (i think) to pass the `ctx` as arguments for the `method`, since it is an object.

gcandal: It must work, because the SDK calls `on_part` a lot. See onPart/onPartP/_onPartOptions/etc

BeeMargarida: @gcandal The problem is that those endpoints are POST, so they send the arguments as body, however the logic endpoint that `getLogic` calls is GET

gcandal: You're right!

gcandal: That's not a problem in practice though, is it?

BeeMargarida: The problem resides in passing a nested object through the URL...i can stringify it, but on the core side it should convert it back to object. Its this, or I'm forgetting something super basic :sweat_smile:

src/js/api/brand.js
```diff
@@ -624,6 +665,39 @@ ripe.Ripe.prototype._runLogicOptions = function(options = {}) {
...
+    const format = options.format === undefined ? this.format : options.format;
+    const method = options.method === undefined ? this.method : options.method;
+    const args = options.args === undefined ? null : options.args;
+    const url = `${this.url}brands/${brand}/models/${model}/logic`;
+    const params = {};
+    if (version !== undefined && version !== null) {
+        params.version = version;
+    }
+    if (format !== undefined && format !== null) {
+        params.format = "js";
```
gcandal, line 684: params.format = format?

gcandal (approved):

joamag (approved):

joamag:
@BeeMargarida It's important to keep the definition order of the methods equal to the definition of the action methods on the server side

https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/controllers/api/brand.py#L224

Should match the corresponding part on the client ence this fix https://github.com/ripe-tech/ripe-sdk/pull/224/commits/56cb024adb273dfb07a752e17356a4068d39f19c

PR #226, opened 2020-12-07
Small fix null brand and model when validating the models

3rdvision:
| - | - |
| --- | --- |
| Issue | The request URL when doing validateModelP will have null brand and model because `_getQueryOptions` will set null brand and model when doing [this](https://github.com/ripe-tech/ripe-sdk/blob/dbc04498e444ed443fb69aabdcb7b7578bced47a/src/js/api/brand.js#L699). image |
| Decisions | Fix by setting brand and model vars before they're set to to null in the options. |


joamag:
Closing this in favor of https://github.com/ripe-tech/ripe-sdk/commit/0bc508a34546df5bc49042e193843539b39b0580

PR #227, opened 2020-12-14
#45: Configurator error on destroy

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk-components-vue/issues/45 |
| Dependencies | -- |
| Decisions | Due to the async nature of the configurator and image execution, it is possible to initiate the destruction of a component that uses the configurator and images while they are updating their state. This would result in the `DOM element` not existing anymore and the components would try to access them, resulting in errors. The solution was returning the functions if the `element` was not defined.  |
| Animated GIF | Showing no errors when initiating and destroying components rapidly: destroy_configurator|

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -624,11 +624,13 @@ ripe.ConfiguratorPrc.prototype.changeFrame = async function(frame, options = {})
  * @param {Object} options Set of optional parameters to adjust the highlighting.
  */
 ripe.ConfiguratorPrc.prototype.highlight = function(part, options = {}) {
+    // in case the element is no longer available (possible due to async
+    // nature of execution) returns the control flow immediately
+    if (!this.element) return;
+
     // verifiers if masks are meant to be used for the current model
     // and if that's not the case returns immediately
-    if (!this.useMasks) {
-        return;
-    }
+    if (!this.useMasks) return;
 
     // captures the current context to be used by clojure callbacks
     const self = this;
@@ -694,13 +696,13 @@ ripe.ConfiguratorPrc.prototype.highlight = function(part, options = {}) {
  * @param {Object} options Set of optional parameters to adjust the lowlighting.
  */
 ripe.ConfiguratorPrc.prototype.lowlight = function(options) {
+    // in case the element is no longer available (possible due to async
+    // nature of execution) returns the control flow immediately
     if (!this.element) return;
 
-    // verifiers if masks are meant to be used for the current model
+    // verifies if masks are meant to be used for the current model
     // and if that's not the case returns immediately
-    if (!this.useMasks) {
-        return;
-    }
+    if (!this.useMasks) return;
 
     // retrieves the reference to the current front mask and removes
     // the highlight associated classes from it and the configurator
@@ -794,9 +796,11 @@ ripe.ConfiguratorPrc.prototype.disableMasks = function() {
  * @private
  */
 ripe.ConfiguratorPrc.prototype._initLayout = function() {
+    // in case the element is no longer available (possible due to async
+    // nature of execution) returns the control flow immediately
     if (!this.element) return;
 
-    // clears the elements children
+    // clears the elements children by iterating over them
     while (this.element.firstChild) {
         this.element.removeChild(this.element.firstChild);
     }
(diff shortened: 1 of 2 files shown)


gcandal:
Can you post the sequence of SDK method calls that leads to this error? The reason I'm asking is that:
- If the error is raised in sync methods, then there is nothing to fix because you aren't supposed to call methods on a configurator after you've called unbind on it
- If the error is raised because you've called an async method and only unbinded the configurator afterwards, then the async method you called should have been cancelled

BeeMargarida:
> Can you post the sequence of SDK method calls that leads to this error? The reason I'm asking is that:
> 
>     * If the error is raised in sync methods, then there is nothing to fix because you aren't supposed to call methods on a configurator after you've called unbind on it
> 
>     * If the error is raised because you've called an async method and only unbinded the configurator afterwards, then the async method you called should have been cancelled

I believe this in the storybook is the extreme, since it is not supposed to destroy and initiate components this way in normal applications. The problems in the configurator consist of several methods, for example the `_preload` function and its methods like `_mark`, which run after a delay. The configurator may have been destroyed  but this method will later try to access this element. 

https://github.com/ripe-tech/ripe-sdk/blob/b2fe273b15b3de28845c710874ef6f046511b654/src/js/visual/configurator-prc.js#L1313

It would result in an error like:
```
Uncaught (in promise) TypeError: this.element is null
    mark configurator-prc.js:1305
    render configurator-prc.js:1351
    promise callback*render configurator-prc.js:1351
    render configurator-prc.js:1353
    render configurator-prc.js:1353
    result configurator-prc.js:1369
    setTimeout handler*./node_modules/ripe-sdk/src/js/visual/configurator-prc.js/ripe.ConfiguratorPrc.prototype._preload/result< configurator-prc.js:1367
    _preload configurator-prc.js:1276
    update configurator-prc.js:247
    _update ripe.js:1256
    update ripe.js:1315
    setPart ripe.js:716
    _callee2$ ripe-pickers.vue:185
    Babel 8
    VueJS 3
configurator-prc.js:1305

```

Another example is the `_updateConfig`, which is called in `init` but is async and does not have an await: https://github.com/ripe-tech/ripe-sdk/blob/b2fe273b15b3de28845c710874ef6f046511b654/src/js/visual/configurator-prc.js#L93

It also happened once in the image `update` method (this is the one that is stranger), which in fact already had the check later in the method https://github.com/ripe-tech/ripe-sdk/blob/b2fe273b15b3de28845c710874ef6f046511b654/src/js/visual/image.js#L408

gcandal:
@BeeMargarida please check:
- if using https://vuejs.org/v2/api/#beforeDestroy to unbind the configurator prevents the problem from happening
- if this is unit-testable

My whole question about this: is this reproducible if you use the SDK correctly by only destroying the HTML element _after_  unbind has completed?

BeeMargarida:
> @BeeMargarida please check:
> 
>     * if using https://vuejs.org/v2/api/#beforeDestroy to unbind the configurator prevents the problem from happening
> 
>     * if this is unit-testable
> 
> 
> My whole question about this: is this reproducible if you use the SDK correctly by only destroying the HTML element _after_ unbind has completed?

@gcandal So, I found a simpler and more "scoped" solution. There are 2 major problems:
* The `destroyed` (same for `beforeDestroy`) can happen before the `bindConfigurator` call, where the `destroyed` call ends and the configurator DOM element does not exist, but it will implement the bindConfigurator with the configurator `$ref` as `undefined`. <br> The **solution** I propose is this, both in `configurator` and `image`: in `ripe-sdk` return if the element is not defined and in `ripe-configurator` and `ripe-image` verify if the result from the bind operation is defined. Or only this last part in the sdk-components.

* The `render` and the `_drawMask` methods in `configurator-prc` run after a delay. Both these functions call others that access the `element`, which may already have been destroyed.  <br> One **solution** would be to verify if the element is undefined in these methods and return. Other would be to save this as promises and cancel them in the `cancel` method, similar to the `loadedCallback` in the `image` [here](https://github.com/ripe-tech/ripe-sdk/blob/eb8eca0273627e6b3f01a3b1cc5636a2ea936038/src/js/visual/image.js#L461) and [here](https://github.com/ripe-tech/ripe-sdk/blob/eb8eca0273627e6b3f01a3b1cc5636a2ea936038/src/js/visual/image.js#L545). I'm not sure this last solution is possible.

gcandal (approved):

joamag (approved):

PR #228, opened 2020-12-15
DataJ problem in _runLogicOptions

BeeMargarida:
| - | - |
| --- | --- |
| Issue | The problem mentioned in the daily, that when passing the context or other options as `data` to the `runLogicP` method, that data does not reach `ripe-core`. |
| Dependencies | -- |
| Decisions | The problem was that both fields `data` and `dataJ` where being passed to yonius. Yonius ignores the `dataJ` if `data` is defined, and adds the `data` to the URL as parameters (as expected), instead of sending it in the request body <br> https://github.com/hivesolutions/yonius/blob/b952ad890b82123009b735e3cdc8c467c9b195b8/js/api/api.js#L119 <br> It is necessary to delete the `data` field when creating the request. Tested by making request, the data is now reaching `ripe-core` and being passed to the method in the model logic script. |
| Animated GIF | -- |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -770,6 +770,7 @@ ripe.Ripe.prototype._runLogicOptions = function(options = {}) {
     if (version !== undefined && version !== null) {
         params.version = version;
     }
+    delete options.data;
     return Object.assign(options, {
         url: url,
         method: "POST",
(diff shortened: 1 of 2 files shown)


gcandal:
Should we just rename `options.data` to `options.dataJ`?

Also, small note: yonius is not used in RIPE SDK but this: https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/api.js#L289

gcandal:
@BeeMargarida please write a unit test for this. I don't know which brands do the unit tests use, but it must either `swear` or `dummy`.

gcandal (approved):

gcandal:
It seems we have no uses of `runLogicP` with `data` so this breaking change should be ok.

joamag (approved):

PR #234, opened 2021-01-27
fix: configurator cancel signature error

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Originated from https://github.com/ripe-tech/ripe-twitch-ui/pull/54 (read description) |
| Dependencies | -- |
| Decisions | From what I understood, the configuration cancel operation proceeds if the signatures don't match, since this means that there is an update process happening. However, the signature assignment was happening earlier in the update process and, when trying to cancel it would not work, since the signatures would match, even with the `_loadFrame` and preload happening. <br><br>By moving the signature assignment below, the signatures will only match when all promises are finished, and the cancel operation will successfully be performed when the `_preload` action is in the process (by running the `_finalize`).<br><br> The reasoning behind this came from a problem where the configurator was canceled in the middle of an update. This caused the SDK to get stuck when updating its children in the code below https://github.com/ripe-tech/ripe-sdk/blob/767dfd77d3635b45e611462bfb95950d96bc205b/src/js/base/ripe.js#L1294 It seems the update gets stuck on the `preloadPromise` of the configurator, which was not canceled by `cancel` called before https://github.com/ripe-tech/ripe-sdk/blob/767dfd77d3635b45e611462bfb95950d96bc205b/src/js/visual/configurator-prc.js#L269 The configuration cancelation never passed by the `_finalize` step, canceling all the render operations, even when it was defined. <br><br> With this fix, after several tries on my part by spamming the "customization" part of the twitch journey, I couldn't not reproduce the error. @3rdvision also tested and only reproduced it with "ridiculous" spamming. In the last gifs there are examples of this spamming, and it works. I don't know if this fixes the problem. |
| Animated GIF | Problem: <br> change_stuff_bug<br> Now: <br>fix-sdk-journey<br>spam-sdk-fixed spam-sdk-bug|

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -286,7 +286,7 @@ ripe.ConfiguratorPrc.prototype.update = async function(state, options = {}) {
  * @param {Object} options Set of optional parameters to adjust the Configurator.
  */
 ripe.ConfiguratorPrc.prototype.cancel = async function(options = {}) {
-    if (this._buildSignature() === this.signature || "") return false;
+    if (this._buildSignature() !== (this.signature || "")) return false;
     if (this._finalize) this._finalize({ canceled: true });
     return true;
 };
@@ -1310,6 +1310,7 @@ ripe.ConfiguratorPrc.prototype._preload = async function(useChain) {
 
             // finalizes the promise by resolving it with
             // the parameter that was just received (final result)
+            console.log("configurator _finalize");
             resolve(result);
         };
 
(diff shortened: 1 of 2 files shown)


gcandal:
@joamag we'll have to take a careful look at this together with @BeeMargarida

@BeeMargarida in the meantime, if you got any free time, please assess if it's feasible to use the `pre_update`/`update`/`post_update` events to disable the ripe-twitch navigation buttons to avoid bumping into this issue.

joamag:
I believe this fix needs to be improved to address the double nature of the signature, prevent double loading, and prevent delayed canceling.
We need to create signature and loadedSignature.

joamag:
@BeeMargarida @gcandal I've changed the solution (not tested at all) but in theory, it should work

joamag:
@BeeMargarida can you please test it extensively as you did with the other solution?

BeeMargarida:
> @BeeMargarida can you please test it extensively as you did with the other solution?

I seems to be working, even in the more stressing situations :+1:

joamag (approved):

joamag:
@BeeMargarida thanks!

PR #32, opened 2018-03-19
fixed initials update and documentation

acaldas:
https://github.com/ripe-tech/ripe-core/issues/3092

screen shot 2018-03-19 at 17 22 08

PR #235, opened 2021-01-28
#32: Type definitions

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/32 |
| Dependencies | -- |
| Decisions | 1. Create a type structure for type definitions used by TypeScript projects like `RIPE Twitch` server-side. `RIPE Twitch` uses this SDK to import an order into the `RIPE Core` ecosystem.<br>2. Ran pretty and cleanup which changed some other files but pretty minor changes |

### DIFF
--- a/src/python/ripe_demo/static/js/main.js
+++ b/src/python/ripe_demo/static/js/main.js
@@ -4,7 +4,7 @@
  */
 var FACES = ["side", "top", "front"];
 
-window.onload = function() {
+window.onload = function () {
     var element = document.getElementById("configurator");
     var _body = document.querySelector("body");
     var url = _body.dataset.url || "https://sandbox.platforme.com/api/";
@@ -37,7 +37,7 @@ window.onload = function() {
         guessUrl: guessUrl
     });
 
-    var randomize = async function() {
+    var randomize = async function () {
         var parts = [];
         for (var key in partsMap) {
             var triplets = partsMap[key];
@@ -48,7 +48,7 @@ window.onload = function() {
         await ripe.setParts(parts, true, { partEvents: false });
     };
 
-    var unique = function() {
+    var unique = function () {
         var count = 1;
         for (var key in partsMap) {
             var triplets = partsMap[key];
@@ -57,7 +57,7 @@ window.onload = function() {
         return count;
     };
 
-    var beautify = function(value) {
+    var beautify = function (value) {
         var buffer = [];
         var parts = value.split("_");
         for (var index = 0; index < parts.length; index++) {
@@ -67,7 +67,7 @@ window.onload = function() {
         return buffer.join(" ");
     };
 
-    var bestFace = function(config) {
+    var bestFace = function (config) {
         var faces = config.faces || [];
         var bestFace = null;
         for (var index = 0; index < FACES.length; index++) {
@@ -84,7 +84,7 @@ window.onload = function() {
         return faces.length > 0 ? faces[0] : null;
     };
 
-    var init = function(instance) {
+    var init = function (instance) {
         initBase(instance);
         initHeader(instance);
         initOAuth(instance);
@@ -92,10 +92,10 @@ window.onload = function() {
         initInitials(instance);
     };
 
-    var initBase = function() {
+    var initBase = function () {
         // registers for the key down event on the global document element
         // to listen to some of the key strokes (global operations)
-        document.addEventListener("keydown", async function(event) {
+        document.addEventListener("keydown", async function (event) {
             if (event.ctrlKey && event.keyCode === 90) {
                 await ripe.undo();
             }
@@ -106,25 +106,25 @@ window.onload = function() {
         });
     };
 
-    var initHeader = function() {
+    var initHeader = function () {
         var setPart = document.getElementById("set-part");
         var setMessage = document.getElementById("set-message");
         var getPrice = document.getElementById("get-price");
         var getCombinations = document.getElementById("get-combinations");
 
         setPart &&
-            setPart.addEventListener("click", function() {
+            setPart.addEventListener("click", function () {
                 randomize();
             });
 
         setMessage &&
-            setMessage.addEventListener("click", function() {
+            setMessage.addEventListener("click", function () {
                 alert("Not implemented");
             });
 
         getPrice &&
-            getPrice.addEventListener("click", function() {
-                ripe.getPrice(function(value) {
+            getPrice.addEventListener("click", function () {
+                ripe.getPrice(function (value) {
                     if (value) {
                         alert(String(value.total.price_final) + " " + value.total.currency);
                     } else {
@@ -134,8 +134,8 @@ window.onload = function() {
             });
 
         getCombinations &&
-            getCombinations.addEventListener("click", function() {
-                ripe.getCombinations(function(combinations) {
+            getCombinations.addEventListener("click", function () {
+                ripe.getCombinations(function (combinations) {
                     alert(
                         "There are <strong>" +
                             String(combinations.length.formatMoney(0)) +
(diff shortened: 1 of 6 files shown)


types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
+type Part = {
+    name: string;
+    material: string;
+    color: string;
+}
+
+type Spec = {
+    brand: string;
+    model: string;
+    parts: Array<Part>;
```
gcandal, line 10: Use the other syntax: `Part[]` or something like that

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
+type Part = {
+    name: string;
+    material: string;
+    color: string;
```
gcandal, line 4: There are optional parts when this probably might be optional. Check the codebase to understand if that's the case

Maybe ask @Benedito0 or @joaofsl for help on debugging this on RIPE White and take a look inside the SDK state for models with optional parts.

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    parts: Array<Part>;
+    initials: string;
+    engraving: string;
+    initialsExtra: object;
+    variant?: string;
+    version?: string;
+    description?: string;
+};
+
+type ProductionOrderOptions = {
```
gcandal, line 19:
```suggestion
type ImportOrderOptions = {
```

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    productId?: string;
+    currency?: string;
+    country?: string;
+    meta?: Array<string>;
+}
+
+export declare class RipeAPI {
+    constructor(options?: any);
+    authKeyP(key: string, options?: object): Promise<void>;
+    importOrderP(ffOrderId: string, options?: ProductionOrderOptions, callback?: Function): Promise<void>;
```
gcandal, line 40: `callback` is not used, delete it

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    _queryToSpec(query: string): Spec;
+}
```
gcandal, line 42: Fix this newline issue with `cleanup .`

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    engraving?: string;
+    initialsExtra?: object;
+    gender?: string;
+    size?: number;
+    pending?: boolean;
+    notify?: boolean;
+    productId?: string;
+    currency?: string;
+    country?: string;
+    meta?: Array<string>;
```
gcandal, line 34: Isn't this an object?

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    variant?: string;
+    parts?: Array<Part>;
+    initials?: string;
+    engraving?: string;
+    initialsExtra?: object;
+    gender?: string;
+    size?: number;
+    pending?: boolean;
+    notify?: boolean;
+    productId?: string;
```
gcandal, line 31: number? not sure

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+type ProductionOrderOptions = {
+    brand?: string;
+    model?: string;
+    variant?: string;
+    parts?: Array<Part>;
+    initials?: string;
+    engraving?: string;
+    initialsExtra?: object;
```
gcandal, line 26: type this further

types/base/api.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    color: string;
+}
+
+type Spec = {
+    brand: string;
+    model: string;
+    parts: Array<Part>;
+    initials: string;
+    engraving: string;
+    initialsExtra: object;
```
gcandal, line 13: type this further

gcandal (changes requested):

types/base/api.d.ts
joao-conde: Does not seem like it:
https://github.com/ripe-tech/ripe-sdk/blob/09f547ab354bfad8fb302bacc69cd2223770f988/src/js/api/order.js#L653

types/base/api.d.ts
joao-conde: Weird, cleanup is not analyzing the `types` folder.

joao-conde: Neither `ts` files. I think the extensions are missing in the `admin-scripts` package.

gcandal: I guess you can consider a PR for admin-scripts then.

In any case, you can obviously just add the newlines manually to move on

types/base/api.d.ts
joao-conde: Im confused, i see some string instances:
- https://github.com/ripe-tech/plugin-farfetch/blob/2cb3979de84ad7cbd7f5b00875384a141271c837/static/themes/farfetch/js/steps/mkfbpixel.part.js#L54
- https://github.com/ripe-tech/plugin-farfetch/blob/2cb3979de84ad7cbd7f5b00875384a141271c837/static/themes/farfetch/js/steps/mkfbpixel.part.js#L54

But I see in Core this cast to int:
- https://github.com/ripe-tech/plugin-farfetch/blob/2cb3979de84ad7cbd7f5b00875384a141271c837/static/themes/farfetch/js/steps/mkfbpixel.part.js#L54

joao-conde: String seems the "better safe than sorry" approach

gcandal: It's a `number`: https://github.com/ripe-tech/ripe-core/search?q=product_id

types/base/api.d.ts
joao-conde: I have tried importing an order with just part name and no material and no color, and with name and material but no color and both test cases failed. I think they are mandatory but I will check with @Benedito0  and @joaofsl  first.

types/base/api.d.ts
```diff
@@ -0,0 +1,57 @@
...
+    method?: string;
+    params?: Record<string, unknown>,
+    headers?: Record<string, unknown>,
+    auth?: boolean
+}
+
+export declare class RipeAPI {
+    key?: string;
+
+    constructor(options?: any);
```
gcandal, line 53: For some reason we're using `unknown` everywhere so stick to it.

types/base/api.d.ts
gcandal: Okay just call it `partTriplet` then.

gcandal (changes requested):

types/base/api.d.ts
joao-conde: My bad. `unknow` is better because when using it forces the user to cast it to the desired type.

src/python/ripe_demo/static/js/main.js
```diff
@@ -4,7 +4,7 @@
  */
 var FACES = ["side", "top", "front"];
 
-window.onload = function() {
+window.onload = function () {
```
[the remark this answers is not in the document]
joao-conde: Should I make a new PR with npm run pretty and cleanup ran?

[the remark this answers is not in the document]
gcandal: You can just revert this file (`git checkout master -- src/python/ripe_demo/static/js/main.js`) and everything is 👍

gcandal (approved):

types/base/api.d.ts
```diff
@@ -0,0 +1,57 @@
...
+    initials: string;
+    engraving: string;
+    initialsExtra: InitialsExtra;
+    variant?: string;
+    version?: string;
+    description?: string;
+};
+
+type ImportOrderOptions = {
+    brand?: string;
```
joamag, line 25: some of these fields should not be considered optional. Please try to investigate

joamag: one obvious is the `model`

joamag (changes requested):

types/base/api.d.ts
gcandal: @joamag I believe they're optional because they fallback to the ones that are configured in RIPE SDK.

gcandal: image

joamag: ahh ok, you're right

joamag (approved):

PR #236, opened 2021-01-29
#32: Order and related types

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/32 |
| Dependencies | -- |
| Decisions | In RIPE Twitch, `importOrderP`, will be used when a Twitch order is paid. `getOrdersP` and `deleteOrderP` will be used for testing purposes. Additionally, having this further typed serves as documentation of the `ripe-sdk` itself. |


types/api/order.d.ts
```diff
@@ -0,0 +1,159 @@
...
+    brand_t?: string;
+    retailer_t?: string;
+    vendor_t?: string;
+    name?: string;
+    justification?: string;
+    timestamp: number;
+    email: string;
+};
+
+export type Attachment = {
```
gcandal, line 19: At least name and kind are mandatory: https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/models/attachment.py#L57

I know it's a pain but it's worse if we do this later when we're working on something unrelated, so please double-check the "optionality" for other models.

joao-conde: Ah, well, I was checking on the `appier.field` declarations and since `mandatory=False` I thought they were optional. Forgot to check the validate. Thank you

types/api/order.d.ts
```diff
@@ -0,0 +1,187 @@
...
+    brand_t?: string;
+    retailer_t?: string;
+    vendor_t?: string;
+    name: string;
+    justification?: string;
+    timestamp: number;
+    email: string;
+}
+
+export type Attachment = {
```
gcandal, line 19: Since this is its own separate model then move it to its own file.

As a rule of thumb, the files for types should be 1-1 with the files of the models.

types/api/order.d.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export enum Gender {
+    male = "male",
+    female = "female",
+    kids = "kids",
+    unknown = "unknown"
+}
+
+export type Order = {
```
gcandal, line 132: Try ordering all these types from "outer" (Order) to "inner" (ComponentPrice/TotalPrice)

gcandal (changes requested):

joao-conde:
Assigning to me as I check if I can use also `getOrders`

types/base/api.d.ts
```diff
@@ -1,15 +1,19 @@
...
+export type Part = {
     name: string;
     material: string;
     color: string;
+    hidden?: boolean;
+    optional?: boolean;
 }
 
-type InitialsExtra = Record<string, {
+export  type InitialsExtra = Record<string, {
```
gcandal, line 11: Doesn't prettier fix this?

```suggestion
export type InitialsExtra = Record<string, {
```

types/base/api.d.ts
```diff
@@ -39,19 +43,30 @@ type ImportOrderOptions = {
...
+
 export declare class RipeAPI {
     key?: string;
 
     constructor(options?: unknown);
     authKeyP(key: string, options?: RequestOptions): Promise<void>;
-    importOrderP(ffOrderId: string, options?: ImportOrderOptions): Promise<void>;
+    importOrderP(ffOrderId: string, options?: ImportOrderOptions): Promise<Order>;
+    getOrdersP(options?: GetRequestOptions): Promise<Order[]>;
+    deleteOrderP(number: number, options?: RequestOptions): Promise<string|null>;
```
gcandal, line 70: When does it return `null`? Can't you `string?`?

gcandal (changes requested):

types/base/api.d.ts
joao-conde: Apparently, `prettier` was not analyzing types, I fixed `package.json` script 👍

types/base/api.d.ts
joao-conde: In return types it looks like you can't use nullable:
image
Furthermore, it says this is either `string` or `null`, never `undefined`.

Looks like the return is wrong regardless as it is not a string. I need to take a deeper look.

types/base/api.d.ts
gcandal: @joao-conde check if that fix needs to be ported to other repositories

types/base/api.d.ts
```diff
@@ -1,57 +1,75 @@
...
+};
+
+export declare class RipeAPI {
+    key?: string;
+
+    constructor(options?: unknown);
+    authKeyP(key: string, options?: RequestOptions): Promise<void>;
+    importOrderP(ffOrderId: string, options?: ImportOrderOptions): Promise<Order>;
+    getOrdersP(options?: GetRequestOptions): Promise<Order[]>;
+    deleteOrderP(number: number, options?: RequestOptions): Promise<XMLHttpRequest | null>;
```
gcandal, line 73: Even though it's true that it returns `{ success: true }` or something like that, that's just a coincidence and has no semantic meaning so the type should be null.

gcandal (approved):

types/base/api.d.ts
joao-conde: But the return sequence is the following:

- https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/api/order.js#L129
- https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/api/order.js#L109
- https://github.com/ripe-tech/ripe-sdk/blob/29fafa0b52866dd383d868491a20752282e3ebd3/src/js/base/api.js#L266 OR  https://github.com/ripe-tech/ripe-sdk/blob/29fafa0b52866dd383d868491a20752282e3ebd3/src/js/base/api.js#L271
- From there it's going to be either `null` or the result of `_requestURLFetch` or `_requestURLLegacy`
- `_requestURLFetch` seems null return as well, so its `null` or `_requestURLLegacy`
- `_requestURLLegacy` returns a `const request = new XMLHttpRequest();`
- So seems like `XMLHttpRequest | null` no?

joao-conde: Ah, so you are saying that the `XMLHttpRequest` is returned so that we can await it but it's irrelevant as a type?

gcandal: yes

types/api/order.d.ts
```diff
@@ -0,0 +1,178 @@
...
+    customer?: string;
+    ambassador?: string;
+    store?: string;
+    number_s?: string;
+    channel?: string;
+    delivery_d?: number;
+    scale?: string;
+    size_scaled?: number;
+    details?: OrderDetails;
+};
```
joamag, line 61: @gcandal We should make these types open so that if new fields exist they don't break, how can we do that?

joamag (changes requested):

types/api/order.d.ts
joao-conde: We could do so as such:

```ts
type A = {
    a?: number;
    b?: string;
    [x: string]: any;
}

const obj: A = {
    a: 4,
    b: "3",
    c: 5
}
```

joao-conde: But then we are not fully taking advantage of the type system (I think). Imagining the `Part` type:

```ts
export type Part = {
    name: string;
    material: string;
    color: string;
    hidden?: boolean;
    optional?: boolean;
};
```
The  following would be allowed:
```ts
const part: Part = {
    name: "laces",
    material: "crocodile",
    colour: "yellow" // notice the key "colour" is misspelled
};
```

joamag: Ok let's do it for any type coming from Appier Models

joao-conde: I agree that if we're using types might as well be as strict and complete as possible. In the long run we save time in bugs prevented by the type system and also get really nice "documentation" because parameters like `options` become clearer.

Also, the fix I presented may not be the best but from some quick research of the web and TS documentation it looks like the only one.

gcandal: @joamag do we keep the type open anyway?

joamag: Let's reverse the change, as unpacking a value from JS with more fields does not generate a runtime error

joamag (changes requested):

types/api/order.d.ts
gcandal: image

joamag:
@joao-conde Let's reverse the open types. We've just verified that there's no runtime error when the base object expands in nature.

joao-conde:
> 
> 
> @joao-conde Let's reverse the open types. We've just verified that there's no runtime error when the base object expands in nature.

Already did

joamag (approved):

PR #237, opened 2021-02-01
feat: support string in _parseExtraS

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Originated from https://github.com/ripe-tech/ripe-twitch-ui/pull/71 |
| Dependencies | -- |
| Decisions | The twitch ui uses only initialsExtra to save the initials state. This caused problems when using an URL with only one group in the initials extra field. <br> The `_parseExtraS` function does not support extraS to be a string. However this can happen if the `initials_extra` in the URL comes with only one group, and the `_unpackQuery` function returns the initials_extra as string instead of an array. <br> This caused errors in the parsing, where instead of returning a correct `initialsExtra` object, it returned an object parsed by each letter. <br><br> Tests for `_parseExtraS` and `_queryToSpec` added. |
| Animated GIF | **Without fix**: <br> image <br> **With fix**: <br> image |

### DIFF
--- a/src/js/base/api.js
+++ b/src/js/base/api.js
@@ -1198,6 +1198,7 @@ ripe.Ripe.prototype._partsMToTriplets = function(partsM, sort = true) {
 
 ripe.Ripe.prototype._parseExtraS = function(extraS) {
     const extra = {};
+    if (!Array.isArray(extraS)) extraS = [extraS];
     for (const extraI of extraS) {
         const [name, initials, engraving] = ripe.splitUnescape(extraI, ":", 2);
         extra[name] = {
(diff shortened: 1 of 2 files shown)


gcandal (approved):

gcandal:
@BeeMargarida try and create a test for this using a higher-level function where you can reproduce the error.

joamag:
I'm going to close this one unless a good rationale exists...
Initials Extra should always be a dictionary:
`
{
    "group" : {
          "initials" : "value",
           "engraving": "value"
     }
}
`

joamag:
I'm closing this one the array type should be ensured by the "outside" with proper array "casting" being done before calling the method. This ensures that the Python parse_extra is equal to the Javascript version.

joamag:
Sorry about that but we should be really pedantic about function signatures and avoid multi-type signatures as they are confusing.

PR #238, opened 2021-02-01
fix: support initialsExtra with one group in _queryToSpec

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Originated from https://github.com/ripe-tech/ripe-twitch-ui/pull/71 |
| Dependencies | -- |
| Decisions | The twitch ui uses only initialsExtra to save the initials state. This caused problems when using an URL with only one group in the initials extra field. <br> The `_parseExtraS` function does not support extraS to be a string. However this can happen if the `initials_extra` in the URL comes with only one group, and the `_unpackQuery` function returns the initials_extra as string instead of an array. <br> This caused errors in the parsing, where instead of returning a correct `initialsExtra` object, it returned an object parsed by each letter. <br><br> Based on the previous PR (https://github.com/ripe-tech/ripe-sdk/pull/237#issuecomment-770874298), the solution was moved to `_queryToSpec`, similar to the tuples approach. |
| Animated GIF |image|

### DIFF
--- a/src/js/base/api.js
+++ b/src/js/base/api.js
@@ -1102,6 +1102,7 @@ ripe.Ripe.prototype._queryToSpec = function(query) {
     const engraving = options.engraving || null;
     let initialsExtra = options.initials_extra || [];
     let tuples = options.p || [];
+    initialsExtra = Array.isArray(initialsExtra) ? initialsExtra : [initialsExtra];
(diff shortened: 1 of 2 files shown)


gcandal (approved):

joamag (approved):

joamag:
@BeeMargarida good fix 👍

PR #240, opened 2021-02-05
feat: localeModel

gcandal:
| - | - |
| --- | --- |
| Decisions | Ported from https://github.com/ripe-tech/ripe-commons-pluginus/blob/986588ada4e104ac273ed9d7f35c0a1dca721203/js/locale/model-locale-resolver.js#L102 which in turn is based on https://github.com/ripe-tech/ripe-compose/blob/master/src/ripe_compose/util/interface.py#L510. |

### DIFF
--- a/src/js/base/logic.js
+++ b/src/js/base/logic.js
@@ -230,14 +230,12 @@ ripe.Ripe.prototype.hasFrame = function(frame) {
  * account for proper build locale usage.
  *
  * @param {String} value The base string value to be used in the localization.
- * @param {String} owner The localization owner, which implements:
- *  - `getLocale(): String`
- *  - `toLocale(value: String, defaultValue: String, locale: String, fallback = true)`
- *  - `getSupportedLocales(): Array<String>`
- *  - `hasLocale(value: String, locale: String)`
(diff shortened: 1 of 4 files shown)


src/js/base/logic.js
```diff
@@ -223,3 +223,43 @@ ripe.Ripe.prototype.hasFrame = function(frame) {
...
+ * current model context.
+ */
+ripe.Ripe.prototype.localeModel = function(
+    value,
+    owner,
+    { brand = null, model = null, locale = null, defaultValue = null, fallback = true } = {}
+) {
+    brand = brand === null ? this.brand : brand;
+    model = model === null ? this.model : model;
+    // locale = locale === null ? this.locale : locale;
```
joamag, line 259: remove comment

joamag (changes requested):

src/js/base/logic.js
gcandal: I've kept this so it serves as a reminder that we need to solve the problem between the conflicting locale definition and the locale() function call, as it's blocking this PR.

joamag:
@gcandal just merged this one without properly testing, please make sure the base storage implementation layer works properly

PR #241, opened 2021-02-10
fix: FileTuple prototype fix

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Related to `juice` problem in PR https://github.com/ripe-tech/epir-papyrus/pull/86 |
| Dependencies | -- |
| Decisions | `FileTuple` prototype assignment was also changing the `Array` prototype https://github.com/ripe-tech/ripe-sdk/blob/3096ffb0c7b9ce341f772e8fab974f3bad64ea12/src/js/base/struct.js#L18 which is used by `cheerio`, a `juice` dependency. This was causing weird behaviour in `juice`, resulting in a fatal error. <br> The solution was using Javascript's [`Object.create`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create). |
| Animated GIF | **Before fix** `Array.prototype` before and after FileTuple prototype assigment: <br>image<br>**After fix** `Array.prototype` before and after FileTuple prototype assigment: <br>image|

### DIFF
--- a/src/js/base/struct.js
+++ b/src/js/base/struct.js
@@ -15,7 +15,7 @@ ripe.FileTuple = function(...args) {
     this.push(...args);
 };
 
-ripe.FileTuple.prototype = Array.prototype;
+ripe.FileTuple.prototype = Object.create(Array.prototype);
 ripe.FileTuple.prototype.constructor = ripe.FileTuple;
 
 ripe.FileTuple.fromData = function(data, name = null, mime = null) {


gcandal (approved):

joamag (approved):

joamag:
@BeeMargarida please create a unit test to make sure there's no regression here

BeeMargarida:
> @BeeMargarida please create a unit test to make sure there's no regression here

There are already unit tests for struct.js and file tuples specifically. Do you mean test specifically that the Array.prototype remains equal?

joamag:
> > @BeeMargarida please create a unit test to make sure there's no regression here
> 
> There are already unit tests for struct.js and file tuples specifically. Do you mean test specifically that the Array.prototype remains equal?

yes

PR #244, opened 2021-02-12
fix: tolerate no element on buildSignature

gcandal:


joamag:
@gcandal what's the issue here?

joamag (approved):

PR #245, opened 2021-02-16
#650: Save and later execute changeFrame operations while preloading

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/650 |
| Dependencies | |
| Decisions | **Problem:** When customizing and then clicking in another part, the `changeFrame` call was being executed when the `configurator` was still `preloading`, resulting in the change frame operation being ignored an never executed: https://github.com/ripe-tech/ripe-sdk/blob/dd48403c58cf4f30d091f28cdc9425fd66de7b9a/src/js/visual/configurator-prc.js#L432-L434 <br>**Solution**: Save the operation in an array of pending operations to be executed later when the preloading promise ends. This needs to also happen when the preload is cancelled, since it would result in lost alterations when doing several customizations. |
| Animated GIF | preferred-frame-fix|

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -71,7 +71,7 @@ ripe.ConfiguratorPrc.prototype.init = function() {
     this._finalize = null;
     this._observer = null;
     this._ownerBinds = {};
-    this.pendingOperations = [];
+    this._pending = [];
 
     // registers for the selected part event on the owner
     // so that we can highlight the associated part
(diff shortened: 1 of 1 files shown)


src/js/visual/configurator-prc.js
```diff
@@ -310,9 +315,31 @@ ripe.ConfiguratorPrc.prototype.update = async function(state, options = {}) {
...
+ */
+ripe.ConfiguratorPrc.prototype.flushPending = async function() {
+    while (this.pendingOperations.length > 0) {
+        const operation = this.pendingOperations.shift();
+        const [name, ...args] = operation;
+        switch (name) {
+            case "changeFrame":
+                await this.changeFrame(...args);
+                break;
+            default:
```
gcandal, line 337: Throw on unknown, to avoid silent errors?

src/js/visual/configurator-prc.js
```diff
@@ -428,8 +455,10 @@ ripe.ConfiguratorPrc.prototype.changeFrame = async function(frame, options = {})
     }
 
     // in case the safe mode is enabled and the current configuration is
-    // still under the preloading situation the change frame is ignored
+    // still under the preloading situation the change frame is saved in
+    // and will be executed after the preloading
     if (safe && this.element.classList.contains("preloading")) {
+        this.pendingOperations = [["changeFrame", frame, options]];
```
gcandal, line 461: Maybe use a named tuple (object) instead of an anonymous one.
```suggestion
        this.pendingOperations = [{ operation: "changeFrame", "args": [frame, options] }];
```

gcandal (commented):
Nice one 👍

BeeMargarida:
@gcandal I will leave it still assigned to me, there is some weird behaviour still happening in some specific cases and will explore better tomorrow

joamag (approved):

joamag:
Merging this please expose the possible issues in a video @BeeMargarida

joamag:
@BeeMargarida please check if the problem still happens (some changes were made) and if that's the case please try to investigate the root cause

BeeMargarida:
> @BeeMargarida please check if the problem still happens (some changes were made) and if that's the case please try to investigate the root cause

I added this to the original issue, but will leave it here as well. I'll investigate the cause
There still remains some problems that happen in specific situations (the end of the gif contains the error):
problem-preferred-frame

PR #248, opened 2021-02-17
#768 Add size to setOptions

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/768 |
| Decisions | Add size to setOptions as suggested in https://github.com/ripe-tech/ripe-commons-pluginus/pull/222#issuecomment-780702882 |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -669,7 +669,7 @@ ripe.Ripe.prototype.setOptions = function(options = {}) {
     this.usePrice = this.options.usePrice === undefined ? !this.noPrice : this.options.usePrice;
     this.noDiag = this.options.noDiag === undefined ? false : this.options.noDiag;
     this.useDiag = this.options.useDiag === undefined ? !this.noDiag : this.options.useDiag;
-    this.size = this.options.size || null;
+    this.size = this.options.size || 1000;
 
     // in case the requested format is the "dynamic" lossless one
     // tries to find the best lossless image format taking into account
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -669,7 +669,7 @@ ripe.Ripe.prototype.setOptions = function(options = {}) {
     this.usePrice = this.options.usePrice === undefined ? !this.noPrice : this.options.usePrice;
     this.noDiag = this.options.noDiag === undefined ? false : this.options.noDiag;
     this.useDiag = this.options.useDiag === undefined ? !this.noDiag : this.options.useDiag;
-    this.size = this.options.size || undefined;
+    this.size = this.options.size || null;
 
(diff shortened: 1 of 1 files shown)


src/js/base/ripe.js
```diff
@@ -668,6 +669,7 @@ ripe.Ripe.prototype.setOptions = function(options = {}) {
     this.usePrice = this.options.usePrice === undefined ? !this.noPrice : this.options.usePrice;
     this.noDiag = this.options.noDiag === undefined ? false : this.options.noDiag;
     this.useDiag = this.options.useDiag === undefined ? !this.noDiag : this.options.useDiag;
+    this.size = this.options.size || undefined;
```
gcandal, line 672: Use `null` as fallback, like the others

src/js/base/ripe.js
```diff
@@ -668,6 +669,7 @@ ripe.Ripe.prototype.setOptions = function(options = {}) {
     this.usePrice = this.options.usePrice === undefined ? !this.noPrice : this.options.usePrice;
     this.noDiag = this.options.noDiag === undefined ? false : this.options.noDiag;
     this.useDiag = this.options.useDiag === undefined ? !this.noDiag : this.options.useDiag;
+    this.size = this.options.size || null;
```
gcandal, line 672: Default to `1000`: https://github.com/ripe-tech/ripe-sdk#options

3rdvision: I have this defaulting to null in other places
https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/visual/image.js#L53
https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/visual/configurator-prc.js#L51

should I also default to 1000 instead?

gcandal: No need I believe, they'll just inherit from here.

gcandal (approved):

joamag:
This should be part of the configurator options and not the ripe instance... also we already have resolution change support in ripe white

image

gcandal:
@joamag:

image

image

Even though it's under `Configurator` on ripe-white, it's actually calling the RIPE instance method: `this.$bus.bind("size_change", size => self.ripe.setSize(size));`

Why shouldn't this be part of `setOptions` if `size` is an option?

joamag:
> @joamag:
> 
> image
> 
> image
> 
> Even though it's under `Configurator` on ripe-white, it's actually calling the RIPE instance method: `this.$bus.bind("size_change", size => self.ripe.setSize(size));`
> 
> Why shouldn't this be part of `setOptions` if `size` is an option?

Seems to be a bad decision of the past, let's leave things as is and not extend the bad decision of the past.

PR #249, opened 2021-02-18
#641 Fix setting initialsExtra in root and options of ripe instance

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/641#issuecomment-780589887 |
| Decisions | This bug was introduced at a [refactor](https://github.com/ripe-tech/ripe-white/commit/53473e502528c44efd547b90bf039c7a06267db3#diff-b671d69de830c33eb2f46a7ef178954ff04079bb2baaa93fa568759b45781accR132) after the [fix](https://github.com/ripe-tech/ripe-white/pull/643). This regression exposes an unexpected RIPE SDK behavior - when we do `setInitialsExtra` it sets `ripe.initialsExtra` but not `ripe.options.initialsExtra`. <br><br>On the other hand `config()` sets both `ripe.initialsExtra` and `ripe.options.initialsExtra`, that's why it was working before the refactor when we were doing everything in `config()`.<br><br>Now RIPE White uses `$ripe.initials`, `$ripe.engraving` and `$ripe.initialsExtra` for the `originalState` and `$ripe.options.initials`, `$ripe.options.engraving` and `$ripe.options.initialsExtra` for the `state`. When we reproduce the bug by changing to a different model it will reset the `originalState` correctly but not the `state` (`ripe.options.initialsExtra` is still dirty with old model's personalization). This buttonText showing the old personalization happens only for a short amount of time depending on how fast the JS code runs because it later receives an event with empty personalization which overrides it but the old state always remains there.<br><br> This fix sets both `ripe.initialsExtra` and `ripe.options.initialsExtra` when doing `setInitialsExtra` <br><br> **Note:** [setSize](https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/ripe.js#L1075) already writes to both `this.size` and `this.options.size` so we might want the same effect for `setInitials` and `setInitialsExtra`. |
| Animated GIF | Peek 2021-02-18 17-09 |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -779,10 +779,18 @@ ripe.Ripe.prototype.setParts = async function(update, events = true, options = {
  * @param {String} engraving The type of engraving to be set.
  * @param {Boolean} events If the events associated with the initials
  * change should be triggered.
+ * @param {Boolean} override If the options value should be override meaning
+ * that further config updates will have this new initials set.
  * @param {Object} params Extra parameters that control the behaviour of
  * the set initials operation.
  */
-ripe.Ripe.prototype.setInitials = async function(initials, engraving, events = true, params = {}) {
+ripe.Ripe.prototype.setInitials = async function(
+    initials,
+    engraving,
+    events = true,
+    override = false,
+    params = {}
+) {
     if (typeof initials === "object") {
         events = engraving === undefined ? true : engraving;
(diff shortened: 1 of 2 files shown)


joamag:
We need to be careful to avoid breaking changes at a core level, I've made updating options optional (using override flag) this is in line with other methods in Ripe and should fix the behaviour.

PR #251, opened 2021-02-19
feat: add import order fields

gcandal:
| - | - |
| --- | --- |
| Decisions | Add missing fields. Some of them for `buildless` orders, some of them for usual orders. Sorted based on [order](https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/models/order.py) fields |


src/js/api/order.js
```diff
@@ -908,14 +908,14 @@ ripe.Ripe.prototype._importOrder = function(ffOrderId, options = {}) {
     const contents = {
         brand: brand,
         model: model,
-        parts: parts
+        parts: parts,
+        size: size
```
joamag, line 912: what you're saying is that size is completely mandatory right?

joamag (changes requested):

joamag (approved):

PR #252, opened 2021-02-19
#44 Parts, color, materials and properties localization support

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/44 |
| Dependencies | -- |
| Decisions | Moved color, material, part and property localization methods logic to ripe-sdk. This methods added are basically [the ones located here](https://github.com/ripe-tech/ripe-commons-pluginus/blob/master/js/locale/model-locale-resolver.js) with a few tweaks, namely they need to receive `owner` as an argument and accept `brand` and `model` as an option parameter |
| Screenshots| **Tests passing:**imagem<br>imagem<br>imagem<br>imagem |


src/js/base/logic.js
```diff
@@ -268,3 +268,269 @@ ripe.Ripe.prototype.localeModel = function(
         fallback: fallback
     });
 };
+
+/**
+ * @ignore
+ */
+ripe.Ripe.prototype._toLocale = function(
```
gcandal, line 275: Delete. This already exists (look 10 lines above for the usage: `ripe.toLocale`)

src/js/base/logic.js
```diff
@@ -268,3 +268,269 @@ ripe.Ripe.prototype.localeModel = function(
...
+        });
+    }
+
+    return values[0];
+};
+
+/**
+ * @ignore
+ */
+ripe.Ripe.prototype._localeModel = function(
```
gcandal, line 328: Delete. This already exists

gcandal (changes requested):

src/js/base/logic.js
NFSS10: But the usage is not the same, this methods was copied (with some minor changes) from commons-pluginus and this seems to handle locale fallback unlike `toLocale` which handles more than that.

I added this method so that the `localeColor`, `localeMaterial`,` localePart` and `localeProperty` maintain their functionallity without messing around too much with the methods, maintaining consistency with this por to ripe-sdk

src/js/base/logic.js
NFSS10: As the above comment, it's not completly the same, as it uses _toLocale and that:
https://github.com/ripe-tech/ripe-sdk/pull/252#discussion_r580112728

gcandal:
Closing until the adaptations we talked about can be done: these methods should mimic https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/models/mixins/locale.py and any "extra" behavior should stay in ripe-commons-pluginus.

gcandal (approved):

joamag (approved):

PR #256, opened 2021-03-12
Pass version along request

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/89 |
| Dependencies | -- |
| Decisions | The Twitch issue won't likely use it and just returns the latest version. However, based on the docstrings, it appears that the version should be passed along the several SDK functions. If the version is defined, the request should be `builds/:name/artifacts/:version` and `builds/:name/artifacts` otherwise|


gcandal:
@joao-conde it's the jsdoc that's wrong, not the endpoint call per see. We always have 1 method = 1 endpoint.

joao-conde:
Oki, going to change the docstrings

PR #263, opened 2021-05-04
Feat: Add promise-based default getter

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Add `getDefaultsP` and bump packages. |


gcandal (approved):

gcandal:
Add a https://github.com/ripe-tech/ripe-sdk/blob/master/CHANGELOG.md entry.

This seems to really be missing, but why do you need it?

joao-conde:
Current approach for CSR-PRC: having the page with the loaded model and change parts that need changing. Problem is for future requests it must be set to the defaults again (because last request may have changed something I do not wish to change).

joamag (approved):

PR #265, opened 2021-05-06
Calling config(), setParts() and setPart() now cascades options to configurators

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | None |
| Dependencies | None |
| Decisions | Below |

When the specified functions are called, they call an `update`, with a specified reason. However, the options which were passed for the functions are not passed on in the update, so a configurator has no way of reading extra information, such as the duration of any event, for example. This caused problems [when attempting to use CSR as an alternative for PRC](https://github.com/ripe-tech/ripe-compose/issues/74) to create static images. 
When changing a configuration, a crossfade event is triggered, in which the CSR seamlessly transitions from one state to the next. For the use case of the issue, we needed the crossfade to be instantaneous, so as to generate a valid frame as soon as possible. Since the functions did not cascade the options, there was no way to explicitly tell the duration (despite the CSR being able to handle the parameter correctly). 

This PR enables this feature, so that we can explicitly call options with duration, which are passed onto the configurator to better deal with this specific issue.

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -722,7 +722,8 @@ ripe.Ripe.prototype.setPart = async function(part, material, color, events = tru
 
     // propagates the state change in the internal structures to the
     // children elements of this Ripe Instance
-    const promise = this.update(undefined, { reason: "set part" });
+    const fullOptions = { ...options, ...{ reason: "set part", noAwaitLayout: true } };
+    const promise = this.update(undefined, fullOptions);
 
     // in case the wait update options is valid (by default) then waits
     // until the update promise is fulfilled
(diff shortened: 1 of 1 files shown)


gcandal:
@veryprofessionaldodo how is `ConfiguratorPrc` able to handle this scenario?

veryprofessionaldodo:
If it is regarding changing the model itself, if I understand the code correctly, the brand and model are passed as mandatory parameters to the `config` function, with only the `options` parameter being optional. The `update` function that is called (that the configurators also use), has explicit parameters that do not take into account any extra information that may be passed through the `config`.

gcandal:
- `duration` is not part of the RIPE API
- Use ConfiguratorPrc.options.duration instead (on init)

PR #266, opened 2021-05-07
#84: Initialls builder in 3db

BeeMargarida:
| - | - |
| --- | --- |
| Dependencies | https://github.com/ripe-tech/ripe-compose/pull/112 ✅  |
| Decisions | - Loads the logic script from the server, defaults to local one. <br> - The local `initialsBuilder` is similar to the one on `ripe-compose`, resulting in the same profiles for the same input. This required the `combinations` method, which is similar to the `itertools.combinations()` python method. The unit tests are similar to the ones on the compose. <br> - `initialsBuilder` signature changed, not receives `initials, engraving, group, viewport, context` (instead of `initials, engraving, element`), similar to the compose. <br><br> **Decisions made:** <br> - Receives both the `context` (static) and a function `getContext` that uses the `group` and possibly other information (position) to build the context. For example in the `?brand=kirkwood&model=alligator_beya_loafer`, the `position` attribute controls the viewport of the image being shown, so its necessary to modify the context based on the current position. To do that, the context can't be static. <br><br> **Note:** Merging this will break all logic with existing initialsBuilder in `retail-vendors-plugins`, since the method signature was changed. |
| Animated GIF | Example for `getContext`: <br> kirkwood_initialsBuilder<br> Using default logic (removed `_initialsBuilder` from `personalization.vue`, using `initialsBuilder` from compose's `base.js` and `_buildProfile` from sdk, passing `getContext` function from `personalization.vue` to `initials-image`) <br>vyner_initialsBuilder<br> Similar to the one before but with 2 groups: <br>air_rev_nitro_initialsBuilder<br> Passing tests: <br> imagem|


BeeMargarida:
CI failing due to https://github.com/Automattic/node-canvas/issues/1806

test/js/visual.js
```diff
@@ -132,4 +132,182 @@ describe("Image", function() {
             assert.strictEqual(image.callbacks, null);
         });
     });
+
+    describe("#_initialsBuilder", function() {
```
gcandal, line 136: @BeeMargarida I've moved these auxiliary methods under `ripe` (instead of `image`) given that they are logical methods and not really about displaying stuff on screen.

Please move the tests as well

src/js/base/ripe.js
```diff
@@ -1972,12 +1972,12 @@ ripe.Ripe.prototype._buildProfiles = function(engraving, profiles, context, sep
     // parses the provided engraving string so that it's possible to iterate
     // over the multiple structured values of it, then adds these same values
     // to the base values passed as argument to the method
-    const engravingProfiles = engraving ? this.owner.parseEngraving(engraving).values : [];
```
gcandal: oops

BeeMargarida: I thinks its correct, I only removed the owner part, which is not necessary now

gcandal (approved):

joamag (approved):

PR #269, opened 2021-05-10
feat: use node-fetch when global.fetch is not defined

BeeMargarida:
| - | - |
| --- | --- |
| Issue | In `epir-papryus` an error occurred when running the deployed version, when the SDK initialization was made. This was due to the fact that `fetch` was not defined: <br>image |
| Dependencies | -- |
| Decisions | Locally this works, but when running `nuxt --standalone`, the `global.fetch` is not defined. After some exploration, the variable `VUE_SSR_CONTEXT` is defined as an empty object in `global` when running nuxt as standalone. |
| Animated GIF | -- |

### DIFF
--- a/src/js/base/compat.js
+++ b/src/js/base/compat.js
@@ -93,6 +93,8 @@ if (
         fetch = window.fetch;
     } else if (typeof global !== "undefined" && typeof global.fetch !== "undefined") {
         fetch = global.fetch;
+    } else if (typeof global !== "undefined") {
+        fetch = require("node-fetch").default;
     } else if (
         // eslint-disable-next-line camelcase
         typeof __webpack_require__ === "undefined" &&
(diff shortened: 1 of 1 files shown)


src/js/base/compat.js
```diff
@@ -93,6 +93,8 @@ if (
         fetch = window.fetch;
     } else if (typeof global !== "undefined" && typeof global.fetch !== "undefined") {
         fetch = global.fetch;
+    } else if (typeof global !== "undefined") {
```
gcandal, line 96: We already add a decision bracket (the one below this) where it lead to the import of node-fetch. What part of that condition is failing?

BeeMargarida: `__webpack_require__` is defined

BeeMargarida:
@gcandal Even by defining the `global.fetch` in papyrus (both in a plugin and before the sdk initialization), it does not work. When it reaches `compat.js`, the `global.fetch` is undefined

src/js/base/compat.js
gcandal: move this as a last branch of the chain of "ifs" and "else ifs". When doing so, the if condition must someshow reflect that we're on a nuxt environmennt (using a var or something)

gcandal (approved):

joamag (approved):

PR #280, opened 2021-06-02
#84: fix initials builder calls

BeeMargarida:
| - | - |
| --- | --- |
| Dependencies | -- |
| Decisions | Initializes `initialsBuilder` with the local implementation, so that if `useInitialsBuilderLogic` is false, it still has it defined. There was a problem that I did not notice (my mistake) with the calls made inside `initialsBuilder`, which needed to be inside the owner `context`. |
| Animated GIF | -- |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -1917,8 +1917,8 @@ ripe.Ripe.prototype._initialsBuilder = function(
     viewport = null,
     context = null
 ) {
-    let profiles = this._generateProfiles(group, viewport);
-    profiles = this._buildProfiles(engraving, profiles, context);
+    let profiles = this.owner._generateProfiles(group, viewport);
+    profiles = this.owner._buildProfiles(engraving, profiles, context);
     return {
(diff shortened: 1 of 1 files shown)


src/js/base/ripe.js
```diff
@@ -1887,7 +1887,7 @@ ripe.Ripe.prototype._loadInitialsBuilder = async function() {
     });
     // eslint-disable-next-line no-eval
     const logicScript = eval(logicScriptText);
-    this.initialsBuilder = logicScript ? logicScript.initialsBuilder : this._initialsBuilder;
+    this.initialsBuilder = logicScript ? logicScript.initialsBuilder : this._initialsBuilder.bind(this);
```
gcandal, line 1890: Shouldn't `logicScript.initialsBuilder` be `bind(this)` as well?

BeeMargarida: Yes, but then I would need to change the ripe-compose PR

gcandal (approved):

PR #286, opened 2021-06-23
#650: Change to preferred frame after loading customization fix

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/650 |
| Dependencies | -- |
| Decisions | • Configurator updates itself by redirecting its update call through its parent (Ripe instance). This allows access control and prevents 2 updates happening in the configurator at the same time, since the configurator does not have checks to prevent this. <br> • By redirecting the update to the parent, it becomes problematic when there are 2 or more configurators listening to the change frame request (e.g. ripe-white). This is because the `update` method uses the `updateCounter` and discards the previous update request if one more recent was made. To fix this, the update requests of the configurators are batched together in `queryUpdate` and only one update request is made with the configurators as the `children` array. This is the simplest way I could find, since the configurators have no information about each other and only ripe can make this logic. |
| Animated GIF | preferred_frame_parent_update |


joamag:
@gcandal @BeeMargarida let's schedule a 30 min call to address this issue

BeeMargarida:
Closing since a better approach was found

PR #292, opened 2021-07-21
#105 Get attachment URL

joaofsl:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-robin-revamp/pull/281#pullrequestreview-711609407 |
| Decisions | Add method to get an attachment URL given its key.|

### DIFF
--- a/src/js/api/order.js
+++ b/src/js/api/order.js
@@ -238,11 +238,6 @@ ripe.Ripe.prototype.attachmentOrderP = function(number, attachmentId, options) {
     });
 };
 
-ripe.Ripe.prototype.getAttachmentOrderUrl = function(key, options = {}) {
-    const params = options.params || {};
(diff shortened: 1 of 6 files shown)


src/js/api/order.js
```diff
@@ -238,6 +238,11 @@ ripe.Ripe.prototype.attachmentOrderP = function(number, attachmentId, options) {
     });
 };
 
+ripe.Ripe.prototype.getAttachmentOrderUrl = function(key, options = {}) {
```
gcandal, line 241: Same approach as `getMeshUrl` and `getLogoUrl`:
- no buildQuery
- _getAttachmentOptions

Also, this is in the wrong file (order.js) and should be in attachment.js (look at the endpoint you're using).

gcandal (approved):

joamag (approved):

PR #303, opened 2021-08-19
fix: support combinations for initials builder of only 1 profile

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Found when implementing initials in ripe-green. |
| Dependencies | -- |
| Decisions | When only when profile was given (which is rarely), the itertools combination would not work, so a check was added, since all combinations of only one element is the element itself. |
| Animated GIF | -- |


gcandal:
Does `ripe.combinations` map 1-1 to some itertools method? If it does, is it behaving the same way?

BeeMargarida:
I verified, the result is the same. I don't know what happened, but I can't replicate the problem I found when I created this PR. I'll close it

Issue #307, opened 2021-08-31
Spec.json Tag to deactivate mask


dsafonso:
## Rationale

Now we have an option on SDK to deactivate the masks and should be nice have  a tag to be add in the model spec to deactivate it.

## Description

Should be possible add a tag example "no_mask" and when this model is on white the masks are deactivated by default.
example:
![image](https://user-images.githubusercontent.com/24550956/131534283-de8b1ec5-16a3-4c79-a464-b6e7601f3bd9.png)
![image](https://user-images.githubusercontent.com/24550956/131534533-7d40bd80-f9a6-4458-ad2f-8e8fa21ef7b5.png)

joamag:
@gcandal Let's do this it should be pretty easy (or not as the config does not exist since the beginning)
@gcandal please explore ideas

dsafonso:
@gcandal Is missing deployment of SDK to test this

dsafonso:
@BeeMargarida Missing update the documentation with this new tag: https://dev.platforme.com/p/specs/build.md

BeeMargarida:
> @BeeMargarida Missing update the documentation with this new tag: https://dev.platforme.com/p/specs/build.md
On it!

dsafonso:
Test with brand `cariuma` model `oca_low` OK :+1:

PR #308, opened 2021-09-02
#307: Disable masks if build contains "no_masks" tag

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/307 |
| Dependencies | -- |
| Decisions | Checks if build contains `no_masks` tag and disables the masks in the configurator. It takes precendence over the options passed to the configurator. |
| Animated GIF | -- |

https://user-images.githubusercontent.com/25725586/131869488-8a8b28b1-1106-4a49-9cec-606c02118a6a.mp4

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -173,6 +173,8 @@ ripe.ConfiguratorPrc.prototype.updateOptions = async function(options, update = 
         options.configAnimate === undefined ? this.configAnimate : options.configAnimate;
     this.viewAnimate = options.viewAnimate === undefined ? this.viewAnimate : options.viewAnimate;
 
+    this.useMasks = this.owner.brand && this.owner.hasTag("no_masks") ? false : this.useMasks;
+
     if (update) await this.update();
 };
 
@@ -1027,6 +1029,8 @@ ripe.ConfiguratorPrc.prototype._updateConfig = async function(animate) {
     // in case the element is no longer available (possible due to async
     // nature of execution) returns the control flow immediately
     if (!this.element) return;
+
+    this.useMasks = this.owner.brand && this.owner.hasTag("no_masks") ? false : this.useMasks;
 
     // sets ready to false to temporarily block
     // update requests while the new config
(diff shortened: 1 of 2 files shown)


BeeMargarida:
Tests failing due to something unrelated to this PR, investigating

src/js/visual/configurator-prc.js
```diff
@@ -173,6 +173,8 @@ ripe.ConfiguratorPrc.prototype.updateOptions = async function(options, update =
         options.configAnimate === undefined ? this.configAnimate : options.configAnimate;
     this.viewAnimate = options.viewAnimate === undefined ? this.viewAnimate : options.viewAnimate;
 
+    this.useMasks = this.owner.brand && this.owner.hasTag("no_masks") ? false : this.useMasks;
```
gcandal, line 176: Why is the `brand` thing needed?

src/js/visual/configurator-prc.js
```diff
@@ -1028,6 +1030,8 @@ ripe.ConfiguratorPrc.prototype._updateConfig = async function(animate) {
     // nature of execution) returns the control flow immediately
     if (!this.element) return;
 
+    this.useMasks = this.owner.brand && this.owner.hasTag("no_masks") ? false : this.useMasks;
```
gcandal, line 1033: Why do you need to do this twice?

src/js/visual/configurator-prc.js
BeeMargarida: To see if the SDK is already configured, or else it will send out errors since `loadedConfig` is null

src/js/visual/configurator-prc.js
BeeMargarida: In case the model changes and therefore the tags might also change

src/js/visual/configurator-prc.js
gcandal: Shouldn't you be look at the `loadedConfig` then? (or isReady)

src/js/visual/configurator-prc.js
gcandal: Sure but wouldn't it suffice to do it here? Isn't `_updateConfig` called the first time?

BeeMargarida: Yes, but for example in the case of ripe-white, the updateOptions sets it to true as default, and the tags should take precedence, I think

gcandal (approved):

joamag (approved):

PR #318, opened 2021-10-08
Added generic bind configurator for PRC and CSR

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/873 |
| Decisions | Currently the `bindConfigurator` method in the SDK equates to generating a PRC configurator. This change allows the configurator to be initialized to be a CSR configurator, while still defaulting to PRC. In the end, no change in behaviour will occur unless it is explicitly stated to initialize a CSR configurator. |


NFSS10:
@veryprofessionaldodo Is this dead? There is no one assigned

veryprofessionaldodo:
@NFSS10 Nice catch, I had forgotten about this to be honest. It was something I did when trying CSR inside RIPE White, and it's still valid. I'll assign it to you, it's quite straightforward and entirely backwards-compatible. The default behaviour is the same as the current one.

src/js/base/ripe.js
```diff
@@ -1195,7 +1195,15 @@ ripe.Ripe.prototype.bindImage = function(element, options = {}) {
...
-    const config = new ripe.Configurator(this, element, options);
+    let config = null;
+    switch (options.type) {
+        case "csr":
+            config = this.bindConfiguratorCsr(element, options);
+            break;
+        case "prc":
+        default:
+            config = this.bindConfiguratorPrc(element, options);
+    }
```
NFSS10, line 1206: SWhouldn't this return instead of assigning the config? I'm a little confused after seeing what `bindConfiguratorPrc` does. Also `bindConfiguratorCsr` doesn't exist

src/js/visual/configurator.js
```diff
@@ -1,18 +0,0 @@
...
-    var base = require("../base");
-    require("./configurator-prc");
-    // eslint-disable-next-line no-redeclare,no-var
-    var ripe = base.ripe;
-}
-
-ripe.Configurator = ripe.ConfiguratorPrc;
-
-ripe.Configurator.prototype = ripe.ConfiguratorPrc.prototype;
-ripe.Configurator.prototype.constructor = ripe.Configurator;
```
NFSS10, line 18: I suppose this was a leftover file that needed to be deleted after the creation of `configurator-prc.js` right?

NFSS10 (changes requested):

NFSS10:
@veryprofessionaldodo Also, don't forget to add the labels and link the issue (if possible) both in the title and in the PR description

src/js/base/ripe.js
veryprofessionaldodo: `bindConfiguratorCsr ` is defined within the external CSR package, in ripe-tech/sandbox. There was the decision to isolate CSR to outside the SDK, and it generates this situation, unfortunately

NFSS10: But about my other question, why there is the `config = ` part? Checking what the functions do, it already does the `BindInteractable(config);`

src/js/visual/configurator.js
veryprofessionaldodo: Previously the configurator class was only a mere redirect to configuratorPRC, so in this PR it serves no purpose, therefore it could be safely removed

src/js/base/ripe.js
veryprofessionaldodo: That's actually a great point! I'm going to try and remove that config, and just return the result of `bindConfiguratorPrc`, see if it works. I might be binding the interactable twice. I'll let you know!

src/js/visual/configurator.js
veryprofessionaldodo: Previously the configurator class was only a mere redirect to configuratorPRC, so in this PR it serves no purpose, therefore it could be safely removed

src/js/base/ripe.js
veryprofessionaldodo: image

Seems like everything is still working as expected! Changing.

src/js/visual/configurator-prc.js
```diff
@@ -12,6 +12,19 @@ if (
...
 
+/**
+ * Binds an PRC Configurator to this Ripe instance.
+ *
+ * @param {Configurator} element The PRC Configurator to be used by the Ripe instance.
+ * @param {Object} options An Object with options to configure the Configurator instance.
+ * @returns {Configurator} The Configurator instance created.
+ */
+ripe.Ripe.prototype.bindConfiguratorPrc = function(element, options = {}) {
+    options = Object.assign({}, { format: this.format }, options);
```
NFSS10, line 23: You are doing this twice, both on `ripe.Ripe.prototype.bindConfigurator` and here.
You should probably remove the one from `ripe.Ripe.prototype.bindConfigurator`

NFSS10 (changes requested):
Also, this: https://github.com/ripe-tech/ripe-sdk/pull/318#issuecomment-976371171

and run pretty, the checks are failling

src/js/visual/configurator-prc.js
veryprofessionaldodo: I'd rather keep that one there, and remove the one from `bindConfiguratorPrc`, since both CSR and PRC could use the same options that are being passed here. Another nice catch, fixing!

NFSS10 (changes requested):
Remove the linting changes of the code not related to this PR If you think that those changed should be commited, create another PR with those changes.

veryprofessionaldodo:
@NFSS10 , weirdly, the only linting suggestions were in the other file. Still, it failed in one of the tests. I'll revert it to it's previous state.

NFSS10:
> @NFSS10 , weirdly, the only linting suggestions were in the other file. Still, it failed in one of the tests. I'll revert it to it's previous state.

Could just be a different problem 🤷‍♂️ 
Either way the ideia is to keep the problems seperated. If the fails are checking before of the changes we made we should fix them, if the checks are failling but it isn't our code, no problem we should investigate why and open a fix in another PR

NFSS10 (approved):

veryprofessionaldodo:
@NFSS10  will do, thanks!

veryprofessionaldodo:
@hugo-gomes Could this be picked up? It's something that's critical for CSR, and it would be a step forward for that to be production-ready.

joamag (approved):

PR #321, opened 2021-10-19
feat: use config animate if update is made after config

BeeMargarida:
| - | - |
| --- | --- |
| Issue | ConfigAnimate was not being used anywhere, making it impossible to disable the animation when first starting the configurator. (The last usage of configAnimate: https://github.com/ripe-tech/ripe-sdk/commit/eae86f176455956d5001d38747c2e579145b52d8#diff-44149027047186f90ca764a3b50eacce3ee815a9f66e73889332689a453b85bd) |
| Dependencies | -- |
| Decisions | - In the configurator `update` method, checks if no signature exists before (meaning that it's a new config), if that is the case, it uses the configAnimate to set the animation (if defined). |
| Animated GIF | **Before:**green-config-before<br>**Now:**<br>Kapture 2021-10-19 at 10 39 56|

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -228,13 +228,24 @@ ripe.ConfiguratorPrc.prototype.update = async function(state, options = {}) {
         const duration = options.duration;
         const preload = options.preload;
 
-        // checks if the parts drawed on the target have
+        // checks if the parts drawn on the target have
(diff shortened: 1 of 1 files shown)


BeeMargarida:
CI failing due to dependency problem, not related to the changes made in this PR. Investigating the problem

joamag (approved):

PR #324, opened 2021-10-27
#923 Add default locale bundle initialization

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/923 |
| Decisions | - Initialize both the SDK locale bundle and the SDK default locale bundle so it's possible to fallback .  |
| Animated GIF | Peek 2021-10-27 09-30 |


src/js/base/ripe.js
```diff
@@ -1623,13 +1623,17 @@ ripe.Ripe.prototype._guessURL = async function() {
...
-    const [globalBundle, sizesBundle] = await Promise.all([globalBundleP, sizesBundleP]);
-    this.addBundle(globalBundle, locale);
-    this.addBundle(sizesBundle, locale);
+ripe.Ripe.prototype._initBundles = async function(locale = this.locale, defaultLocale = "en_us") {
+    const locales = [defaultLocale, locale];
+
+    for (const locale of new Set(locales)) {
+        const globalBundleP = this.localeBundleP(locale, "scales");
+        const sizesBundleP = this.localeBundleP(locale, "sizes");
+        const [globalBundle, sizesBundle] = await Promise.all([globalBundleP, sizesBundleP]);
```
joamag, line 1632: I believe we can make this a parallel set of calls, together with the interaction cycle it is included in. This should bring better performence.

src/js/base/ripe.js
```diff
@@ -1623,13 +1623,17 @@ ripe.Ripe.prototype._guessURL = async function() {
...
  * @ignore
  */
-ripe.Ripe.prototype._initBundles = async function(locale = null, defaultLocale = "en_us") {
-    locale = locale || this.locale || defaultLocale;
-    const globalBundleP = this.localeBundleP(locale, "scales");
-    const sizesBundleP = this.localeBundleP(locale, "sizes");
-    const [globalBundle, sizesBundle] = await Promise.all([globalBundleP, sizesBundleP]);
-    this.addBundle(globalBundle, locale);
-    this.addBundle(sizesBundle, locale);
+ripe.Ripe.prototype._initBundles = async function(locale = this.locale, defaultLocale = "en_us") {
```
joamag, line 1626: I'm not sure you can access this within the default value initialization... At least in Python it would not work

joamag (changes requested):

joamag:
@3rdvision I'm not sure about the overall strategy. Why do we need the default (fallback) locale bundle to be loaded?

3rdvision:
> Why do we need the default (fallback) locale bundle to be loaded?

Because the default fallback locale is not loaded and the [issue](https://github.com/ripe-tech/ripe-white/issues/923) appears when we visit a ripe-white link with a locale that doesn't exist, it first loads the empty locale map for the nonexistent locale and then falls back to the default `en_us` locale which never loaded in the first place.

src/js/base/ripe.js
3rdvision: I believe so, tested locally and is a working solution

joamag (approved):

PR #335, opened 2021-12-17
#276: shipment API

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/276 |
| Dependencies | https://github.com/ripe-tech/ripe-core/pull/4649 |
| Decisions | Create Shipment API. Create methods `getShipments/getShipmentsP` and `createShipment/createShipmentsP`. <br> The params are all documented. While extensive, I feel like external SDK users can have more autonomy this way. It also helps us internally, making teams more autonomous and non-reliant on a specific person's knowledge. |


src/js/api/shipment.js
```diff
@@ -0,0 +1,152 @@
...
+ripe.Ripe.prototype.getShipmentsP = function(options) {
+    return new Promise((resolve, reject) => {
+        this.getShipments(options, (result, isValid, request) => {
+            isValid ? resolve(result) : reject(new ripe.RemoteError(request, null, result));
+        });
+    });
+};
+
+/**
+ * Create a shipment.
```
BeeMargarida, line 66: That is a bit short, please change to something like (same below):
```suggestion
 * Creates a shipment on Ripe Core.

```

test/js/api/shipment.js
```diff
@@ -0,0 +1,80 @@
+const assert = require("assert");
+const config = require("../config");
+const ripe = require("../../../src/js");
+
+describe("ShipmentAPI", function() {
```
BeeMargarida, line 5: also add a test for delete shipment function [inspiration](https://github.com/ripe-tech/ripe-sdk/blob/9fd7461be6bef3b4e479d4a0df8e4b38c3bd6bf6/test/js/api/sku.js#L316)

src/js/api/shipment.js
```diff
@@ -0,0 +1,259 @@
...
+ * Updates a shipment on Ripe Core.
+ *
+ * @param {Number} number The number of the shipment to find by.
+ * @param {Object} options An object with options, such as:
+ *  - 'status' - The shipment status.
+ *  - 'order' - The order associated with this shipment.
+ *  - 'bulk_order' - The bulk order associated with this shipment.
+ *  - 'courier' - The courier for this shipment.
+ *  - 'tracking_number' - The tracking number associated with this shipment.
+ *  - 'tracking_url' - The tracking url associated with this shipment.
```
BeeMargarida, line 207: Please replicate in the other documentation in this file
```suggestion
 *  - 'tracking_url' - The tracking URL associated with this shipment.

```

src/js/api/shipment.js
```diff
@@ -0,0 +1,359 @@
...
+ *  - 'delivery_date' - The date the shipment began.
+ *  - 'shipper' - Shipper's information.
+ *    - 'name' - Shipper's contact name.
+ *    - 'phone' - Shipper's contact phone.
+ *    - 'address' - Shipper's address.
+ *        - 'line' - Shipper's address line.
+ *        - 'city' - Shipper's address city.
+ *        - 'postal_code' - Shipper's postal code.
+ *        - 'state_code' - Shipper's state province code.
+ *        - 'country_code' - Shipper's 'ISO 3166-2' country code.
```
BeeMargarida, line 304: I don't know if you should have this here but leave it vague in the data model in core. This is weird, since you expect a certain structure but have no way of verifying it/forcing it in the data model.

(this is just a comment, no change required)

BeeMargarida (approved):

src/js/api/shipment.js
joao-conde: I understand. My thought process was that this needs to be documented somewhere. For example, Marina as a stakeholder of drop shipping constantly asks me what the format needs to be for certain fields. I thought about documenting it here in the SDK since it will be used by external teams or developers. This way it is documented but it can of course be moved somewhere else accordingly. What do you think @hugo-gomes ?

src/js/api/shipment.js
```diff
@@ -0,0 +1,313 @@
...
+    return this._cacheURL(options.url, options, callback);
+};
+
+/**
+ * Updates a shipment on Ripe Core.
+ *
+ * @param {Number} number The number of the shipment to find by.
+ * @param {Object} options An object with options, such as:
+ *  - 'status' - The shipment status.
+ *  - 'description' - A brief description of this shipment.
```
BeeMargarida, line 244: missing `'orders'`

src/js/api/shipment.js
```diff
@@ -0,0 +1,313 @@
...
+    });
+};
+
+/**
+ * Updates a shipment on Ripe Core.
+ *
+ * @param {Number} number The number of the shipment to find by.
+ * @param {Object} options An object with options, such as:
+ *  - 'status' - The shipment status.
+ *  - 'description' - A brief description of this shipment.
```
BeeMargarida, line 196: missing `'orders'`

test/js/api/shipment.js
```diff
@@ -0,0 +1,339 @@
...
+            const remote = ripe.RipeAPI();
+
+            result = await remote.authAdminP(config.TEST_USERNAME, config.TEST_PASSWORD);
+
+            assert.strictEqual(result.username, config.TEST_USERNAME);
+            assert.notStrictEqual(typeof result.sid, undefined);
+
+            const shipment = await remote.createShipmentP({
+                status: "sent",
+                orders: [4488],
```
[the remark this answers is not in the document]
joao-conde: I am using one used in other tests, particularly from the order tests, so it should always exist.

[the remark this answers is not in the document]
BeeMargarida: not if you run locally, which SDK now supports

src/js/api/shipment.js
```diff
@@ -217,6 +217,7 @@ ripe.Ripe.prototype.createShipmentP = function(options, callback) {
  *    - 'name' - Customer's contact name.
  *    - 'phone' - Customer's contact phone.
  *    - 'address' - Customer's address.
+ *  - 'orders' - A list of RIPE Core order IDs.
```
BeeMargarida, line 220: In some docs you have this in different ordering

test/js/api/shipment.js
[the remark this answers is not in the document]
joao-conde: True, but then I think the order tests need to be updated too. Shouldn't it be done in a separate PR?

src/js/api/shipment.js
joao-conde: I followed the model order. References usually come last from what I can tell.

BeeMargarida: In `createShipment' and `createShipmentP` docs you have the orders docs right after status, just saying they should be equall

joao-conde:
@BeeMargarida  made some modifications (I wasn't passing the payload to the SDK method 🤦‍♂️ ). Please re-review 🙏

src/js/api/shipment.js
```diff
@@ -0,0 +1,319 @@
...
+        url: url,
+        method: "GET",
+        auth: true
+    });
+    options = this._build(options);
+    return this._cacheURL(options.url, options, callback);
+};
+
+/**
+ * Gets an shipment by number.
```
BeeMargarida, line 87:
```suggestion
 * Gets a shipment by number.

```

src/js/api/shipment.js
```diff
@@ -0,0 +1,319 @@
...
+ *  - 'service' - The transportation service used.
+ *  - 'package' - The package used.
+ *  - 'tracking_number' - The tracking number associated with this shipment.
+ *  - 'tracking_url' - The tracking URL associated with this shipment.
+ *  - 'carbon_neutral' - If this shipment should be carbon neutral.
+ *  - 'access_point_delivery' - If this shipment should be sent to an access point (e.g. "no", "optional" or "mandatory").
+ *  - 'pickup' - Pickup information.
+ *    - 'offset' - Time offset for pickup scheduling relative to shipment creation (millis).
+ *    - 'ready_time' - Opening hours for shipper facility (millis).
+ *    - 'close_time' - Closing hours for shipper facility (millis).
```
BeeMargarida, line 119: We usually write the whole word `milliseconds` instead of `millis`. Better to follow the presumed standard we have regarding this

src/js/api/shipment.js
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * Deletes a shipment by number.
+ *
+ * @param {Number} number The number of the shipment to delete.
+ * @param {Object} options An object of options to configure the request.
+ * @param {Function} callback Function with the result of the request.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
+ripe.Ripe.prototype.deleteShipment = function(id, options, callback) {
```
BeeMargarida, line 293: Since it's by number, perhaps change id to number
```suggestion
ripe.Ripe.prototype.deleteShipment = function(number, options, callback) {

```

BeeMargarida (commented):
It would be cool to have types for this, so that at least the information received and returned by the endpoint is easy to understand. (not sayying to do it, since it will not be useful for us now)

hugo-gomes (approved):

joamag (approved):

PR #336, opened 2021-12-23
test: allow config core instance

joao-conde:
| - | - |
| --- | --- |
| Issue | Originated via https://github.com/ripe-tech/ripe-pulse/issues/276 |
| Dependencies | |
| Decisions | To run ripe-sdk tests locally, it is useful to change the core instance used. I kept the same default (SBX). |
| Animated GIF | -- |

### DIFF
--- a/src/python/ripe_demo/static/js/main.js
+++ b/src/python/ripe_demo/static/js/main.js
@@ -2,419 +2,420 @@
  * The complete set of faces that may be used in the configurator
  * as the starting view.
  */
-var FACES = ["side", "top", "front"];
+ var FACES = ["side", "top", "front"];
 
-window.onload = function () {
-    var element = document.getElementById("configurator");
-    var _body = document.querySelector("body");
-    var url = _body.dataset.url || "https://sandbox.platforme.com/api/";
-    var brand = _body.dataset.brand || "swear";
-    var model = _body.dataset.model || "vyner";
-    var variant = _body.dataset.variant || "";
-    var version = _body.dataset.version || null;
-    var format = _body.dataset.format || "lossless";
-    var currency = _body.dataset.currency || "USD";
-    var country = _body.dataset.country || "US";
-    var clientId = _body.dataset.client_id || null;
-    var clientSecret = _body.dataset.client_secret || null;
-    var guess = ["1", "true", "True"].indexOf(_body.dataset.guess) !== -1;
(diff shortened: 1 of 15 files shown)


NFSS10 (changes requested):
Revert the pretty changes and also make sure the tests are ok, some of them check the URL and it's not checking against the TEST_URL

src/python/ripe_demo/static/js/main.js
```diff
@@ -2,419 +2,420 @@
...
-        }
-        if (bestFace) {
-            return bestFace;
-        }
-        return faces.length > 0 ? faces[0] : null;
-    };
-
-    var init = function(instance) {
-        initBase(instance);
-        initHeader(instance);
```
NFSS10, line 89: The changes of this whole file should be reversed

test/js/api/attachment.js
```diff
@@ -7,7 +7,7 @@ describe("AttachmentAPI", function() {
 
     describe("#getAttachmentOrderUrl()", function() {
         it("should be able to generate a simple attachment URL", async () => {
-            const remote = ripe.RipeAPI();
+            const remote = ripe.RipeAPI({ url: config.TEST_URL });
             const result = remote.getAttachmentOrderUrl(123123123);
             assert.strictEqual(result, "https://sandbox.platforme.com/attachments/123123123/data?");
```
NFSS10, line 12: Shouldn't this check against `config.TEST_URL` instead of the hardcoded "https://sandbox.platforme.com"?
I think you should have a `config.TEST_WEB_URL` that it's the url without the `/api/` suffix and use that.

For reference:
https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/ripe.js

NFSS10 (changes requested):

test/js/api/attachment.js
joao-conde: Ye this URL is as you said the `webUrl` so it does not make sense to compare it to `config.TEST_URL`.
Because no web url is given as a configuration variable, this always passes. 

Since my use case and this PR concerned only the change of the API URL, I did not allow for web url configuration. I think its best this way, if need be, by necessity, we can allow web url to be changed, maybe even have a `getRipeOptions` method for the tests to avoid repetition. But I wouldn't do it here.

NFSS10 (approved):

hugo-gomes (approved):

joamag (approved):

PR #338, opened 2022-01-05
feat: text length endpoint

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | - `textLengthP` so that we can make logic similar to what we have for YSL in JS (so that they can use the SDK)<br>- Added `ctx` to the initials builder call|
| Animated GIF | -- |


joamag (approved):

BeeMargarida:
I'll fix the tests

PR #348, opened 2022-01-17
fix: convert image size and width height to ints

BeeMargarida:
| - | - |
| --- | --- |
| Issue | In computers with non-int pixel ratio, this occured (width is float):<br>imagem<br>imagem |
| Dependencies | -- |
| Decisions | - Convert image `size` and `width` and `height` to ints, since compose only supports ints. |
| Animated GIF | -- |


BeeMargarida:
@joamag This is required by YSL to conclude their validation

joamag (approved):

PR #354, opened 2022-02-02
Initials builder ctx message dispatcher

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/builds-static/pull/2039#discussion_r796827127 |
| Decisions | - Add initials builder `ctx.messages` dispatcher handler for build's js logic  |

### DIFF
--- a/src/js/visual/image.js
+++ b/src/js/visual/image.js
@@ -421,7 +421,7 @@ ripe.Image.prototype.update = async function(state, options = {}) {
         });
 
         // if there are message events in initials builder ctx, dispatch them
-        if (initialsSpec?.ctx?.messages?.length) {
+        if (initialsSpec.ctx && initialsSpec.ctx.messages && initialsSpec.ctx.messages.length) {
             for (const [topic, content] of initialsSpec.ctx.messages) {
                 this.owner.trigger("message", topic, content);
             }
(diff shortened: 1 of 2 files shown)


3rdvision:
re-assigning to fix tests

src/js/visual/image.js
```diff
@@ -420,6 +420,13 @@ ripe.Image.prototype.update = async function(state, options = {}) {
             curve: curve
         });
 
+        // if there are message events in initials builder ctx, dispatch them
+        if (initialsSpec.ctx && initialsSpec.ctx.messages && initialsSpec.ctx.messages.length) {
```
joamag, line 424: This should be done when running the initials builder to be aligned with the server side.
I'm not sure this is the "perfect" place to position this code

joamag (changes requested):

src/js/visual/image.js
3rdvision: Initials builder runs here https://github.com/ripe-tech/ripe-sdk/blob/e3bc605b7c243305bd65a7ef1019b303382be09a/src/js/visual/image.js#L349

`initialsSpec` holds the result of running the initials builder. I will move it closer to the call...

joamag (approved):

PR #360, opened 2022-02-25
style: pretty

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | I do not see a reason not to do this albeit there might be one. It is annoying that some files were not linted/prettified. Not only should they be, but we have a huger local git diff unnecessarily after running pretty. No one wants to commit these changes in non-related PRs, so I am opening this one. |


BeeMargarida:
@joao-conde Relative to the javascript file inside src/python, I don't think that is the solution, I think it is related to the differences in setup between prettier and lint. (lint only targets src/js and prettier targets all src). The linting changes made to that file to not match the standards we use.
I think I can fix this setup problem, let me try

Issue #357, opened 2022-02-18
SDK slow response when doing setInitialsExtra


3rdvision:
## Description

SDK takes too long to update images when calling for several `setInitialsExtra`.

In practice ripe-white takes much longer - more than 2x the amount of time - than RIPE API to render an image after user initials requests.

I believe this has to do with SDK's implementation to cancel old requests - it's not working as expected. 

Please watch the videos below.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The image should render what it's typed on white right away, not the first character and then the whole word.  |
| **Observed** | The observed state at the end of the repro steps. |

## Repro Steps

1. Open ripe-white https://ripe-white-now.platforme.com/?context=saint_laurent_hoodie
2. Write something in initials input with more than 1 character like "1234"
2. Watch the image first render an image with the first character "1" and only then "1234" no matter how fast it's typed

### From RIPE White

https://user-images.githubusercontent.com/24736423/154675074-28cfaf4d-8b71-4693-a583-aaf27e82d030.mp4

### From YSL's Website

https://user-images.githubusercontent.com/24736423/154671733-9ef37962-0f57-49fe-8a61-95b511df74c7.mp4

3rdvision:
This can be solved by implementing a debouncer when doing `setInitialsExtra`.

Now this debouncer could be implemented at SDK level and enabled/disabled with a certain timer by default or we just don't implement it and instead pass that responsibility to whoever uses the SDK.

@joamag  I'd say this is a point of discussion in tech decision.

joamag:
> discussion

Yes, let's implement it at the SDK level. On top of that let's canceled previous requests when new ones arise, if needed.

3rdvision:
> On top of that let's canceled previous requests when new ones arise, if needed.

That one is [already implemented](https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/ripe.js#L1368)

Ok, going ahead with the debouncer implementation at SDK level for the `setInitialsExtra`.

luisamd:
@3rdvision Is this story ready for QA? If so, please update the Kanban board 😄

3rdvision:
@msp-platforme Deployed and ready to test at https://ripe-white-now.platforme.com/?context=saint_laurent_hoodie

msp-platforme:
We no longer see the first character being rendered first, and then rendering the rest of the inserted characters.
In that regard, the expectation is met.
[...]

PR #363, opened 2022-02-28
#357 Debouncer for setting the initials

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/357 |
| Decisions | - Use debouncer timer `initialsDebounce=275` for `setInitials` and `setInitialsExtra`. <br> - Adapt tests |

Notes:
* Frontend and backend debouncer strategies were tested and had similar response times.
* Doing a debouncer within `update` was the first approach but failed to be optimal as the event `initials_extra` triggers for an `update` with a null `params.reason` and it's not possible to adapt it without a huge refactor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Add create order bulk methods - [ripe-pulse/#281](https://github.com/ripe-tech/ripe-pulse/issues/281)
+* Add debouncer `initialsDebounce` defaulting to 275 for `setInitials` and `setInitialsExtra` methods - [#357](https://github.com/ripe-tech/ripe-sdk/issues/357)
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -312,6 +312,7 @@ if (ripe.isOAuthPending()) {
 | `usePrice`                | *boolean*          | Enables the fetch price feature every time a new part is set. True by default.                                                                                                             |
 | `useSync`                 | *boolean*          | Enables the part synchronization feature. False by default.                                                                                                                                |
 | `useInitialsBuilderLogic` | *boolean*          | Enables the usage of the client-side initials builder logic defined in the 3DB, instead of the default one. True by default.                                                               |
+| `initialsDebounce`        | *number*           | The debounce time to consider for every initials set request in ms (0 to not debounce). 275 by default.                                                                                    |
 
 ## Browser Support
 
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ if (ripe.isOAuthPending()) {
 | `usePrice`                | *boolean*          | Enables the fetch price feature every time a new part is set. True by default.                                                                                                             |
 | `useSync`                 | *boolean*          | Enables the part synchronization feature. False by default.                                                                                                                                |
 | `useInitialsBuilderLogic` | *boolean*          | Enables the usage of the client-side initials builder logic defined in the 3DB, instead of the default one. True by default.                                                               |
-| `initialsDebounce`        | *number*           | The debounce time to consider for every initials set request in ms (0 to not debounce). 275 by default.                                                                                    |
+| `initialsDebounce`        | *number*           | The debounce time (in milliseconds) for every initials set request in ms. 275 by default and 0 to disable.                                                                                    |
 
 ## Browser Support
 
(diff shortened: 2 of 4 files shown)


BeeMargarida (commented):
@3rdvision Just to check, did you explore a possible solution where a cancel flag is used to stop the update promise before the existence of `loadedCallback`?
For example, setting a `cancelFlag` on `cancel` that would be unset after the cancel request in update. If the flag was true during the update process (for example before `getImageUrl` or `doubleBuffer`) it would return immediately and not complete the whole update

3rdvision:
> Just to check, did you explore a possible solution where a cancel flag is used to stop the update promise before the existence of `loadedCallback`?

I did made an implementation of it but it didn't offer much performance difference. The problem is update will await any previous updates [here](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L1389) after it cancels pending updates [here](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L1383) and there wasn't a significant performance improvement, maybe about from 6.2s to 5.2s while using the debouncer it's about 6.2s to 3.1s.

It's also difficult to distinguish set initials update. In some places we have [reason specifying set initials](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L880) but because we also trigger [initials_extra](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L970) to notify the client that the initials changed after `setInitialsExtra`, we have a [bind for initials_extra](https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/ripe.js#L211) and in other places which will do a setInitials later without including params.reason="set initials_extra" and even thought we could refactor to always track it, from what I've tried, that experience didn't hold significant performance improvements versus this one.

BeeMargarida:
> > Just to check, did you explore a possible solution where a cancel flag is used to stop the update promise before the existence of `loadedCallback`?
> 
> I did made an implementation of it but it didn't offer much performance difference. The problem is update will await any previous updates [here](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L1389) after it cancels pending updates [here](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L1383) and there wasn't a significant performance improvement, maybe about from 6.2s to 5.2s while using the debouncer it's about 6.2s to 3.1s.
> 
> It's also difficult to distinguish set initials update. In some places we have [reason specifying set initials](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L880) but because we also trigger [initials_extra](https://github.com/ripe-tech/ripe-sdk/blob/231277f38c6546d338aa9f34f97121af17826840/src/js/base/ripe.js#L970) to notify the client that the initials changed after `setInitialsExtra`, we have a [bind for initials_extra](https://github.com/ripe-tech/ripe-sdk/blob/master/src/js/base/ripe.js#L211) and in other places which will do a setInitials later without including params.reason="set initials_extra" and even thought we could refactor to always track it, from what I've tried, that experience didn't hold significant performance improvements versus this one.

What I was sugesting was end the updatePromise earlier on cancel, since if the update process did yet not reach the `new Promise.....` that is saved as `loadedCallback` when the cancel is called, the whole update is made. With the approach I mentioned, this does not happen. For example, when writting a whole word, only the last request is completed (with the whole word), instead of the first and then the last.

Example inside `_update` closure in `update`: 
```
...
// verifies if the model currently loaded in the Ripe Instance can
// render the frame to be display and if that's not the case "ignores"
// the current request for update
if (frame && !this.owner.hasFrame(frame)) {
    this.trigger("not_loaded");
    return false;
}

if (this.cancelFlag) return;

// builds the URL of the image using the frame hacking approach
// this should provide us with the new values
const url = this.owner._getImageURL({
    frame: frame,
    ...
});
...

The `cancelFlag` is set as true in the `cancel` method and as false after the cancel call in `update`. If the previous update has not yet reached the assignment of `this._loadedCallback`, the `_update` method would return instead of executing until the end.

3rdvision:
Closing this in favor of a more performant approach https://github.com/ripe-tech/ripe-sdk/pull/366

PR #366, opened 2022-03-02
#363 YSL: Improve update cancellation strategy

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/357 |
| Decisions | - Improve update cancellation strategy by adding a `cancel` event to cancel ongoing update promises |

Note: Although a [debouncer solution](https://github.com/ripe-tech/ripe-sdk/pull/363) could still be implemented, this change is enough to solve the identified performance issue.

### DIFF
--- a/src/js/visual/image.js
+++ b/src/js/visual/image.js
@@ -358,10 +358,7 @@ ripe.Image.prototype.update = async function(state, options = {}) {
               )
             : {};
 
-        if (this.cancelFlag) {
-            console.log("this.cancelFlag is true, returning...");
-            return;
-        }
+        if (this.cancelFlag) return;
 
         // if there are message events in initials builder ctx, dispatches
         // them to the proper message handler (to display message to end user)
@@ -377,11 +374,6 @@ ripe.Image.prototype.update = async function(state, options = {}) {
         if (frame && !this.owner.hasFrame(frame)) {
             this.trigger("not_loaded");
             return false;
-        }
-
-        if (this.cancelFlag) {
-            console.log("this.cancelFlag is true, returning...");
-            return;
         }
 
         // builds the URL of the image using the frame hacking approach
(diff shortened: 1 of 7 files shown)


src/js/visual/image.js
```diff
@@ -511,6 +516,7 @@ ripe.Image.prototype.update = async function(state, options = {}) {
  * instead no cancel logic was executed.
  */
 ripe.Image.prototype.cancel = async function(options = {}) {
+    this._cancellingUpdates = true;
```
BeeMargarida, line 519: newline below

src/js/visual/image.js
```diff
@@ -358,6 +358,8 @@ ripe.Image.prototype.update = async function(state, options = {}) {
               )
             : {};
 
+        if (this._cancellingUpdates) return;
```
BeeMargarida, line 361: Perhaps just add a comment explaining the reason for this line, since these functions are heavily documented and the logic of update is complex

BeeMargarida (commented):
Just doc things :ok_hand: 
Just as a curiosity, is the performance similar to the debouncer solution?

3rdvision:
> Just doc things 👌 Just as a curiosity, is the performance similar to the debouncer solution?

The performance is better because we no longer have to always wait for the debouncer's default 275ms timeout 😄

BeeMargarida (approved):

joao-conde (approved):

3rdvision:
Going to refactor to cancel the `await this.initialsBuilder` instead

joamag:
@3rdvision Adding fast-track to this one as YSL needs this one

joamag (approved):

joamag:
@3rdvision We need to double-check that everything works. But I believe this solution is regenerator-runtime proof

PR #370, opened 2022-03-09
#979: Methods `getSkusCsvP` and `countSkusP`

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/979 |
| Dependencies | https://github.com/ripe-tech/ripe-core/pull/4698 |
| Decisions | - Add methods `getSkusCsvP` and `countSkusP` that return a CSV containing SKU and the number of documents, respectively |
| Animated GIF | -- |

### DIFF
--- a/src/js/api/sku.js
+++ b/src/js/api/sku.js
@@ -79,7 +79,6 @@ ripe.Ripe.prototype.getSkusP = function(options) {
         auth: true
     });
     options = this._build(options);
-    console.log("OPTIONS", options);
     return this._cacheURL(options.url, options, callback);
 };
 
--- a/src/js/api/sku.js
+++ b/src/js/api/sku.js
@@ -55,6 +55,50 @@ ripe.Ripe.prototype.getSkusP = function(options) {
 };
 
 /**
+ * Gets the existing SKUs in a CSV file, according to the provided filtering
+ * strategy as normalized values.
+ *
+ * @param {Object} options An object of options to configure the request, such as:
+ * - 'filters[]' - List of filters that the query will use to, operators such as
+ * ('in', 'not_in', 'like', 'contains'), for instance (eg: 'id:eq:42') would filter by the id that equals to 42.
+ * - 'sort' - List of arguments to sort the results by and which direction
+ * to sort them in (eg: 'id:ascending') would sort by the id attribute in ascending order,
+ * while (eg: 'id:descending')] would do it in descending order.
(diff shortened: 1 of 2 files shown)


src/js/api/sku.js
```diff
@@ -54,6 +54,50 @@ ripe.Ripe.prototype.getSkusP = function(options) {
...
+    callback = typeof options === "function" ? options : callback;
+    options = typeof options === "function" || options === undefined ? {} : options;
+    const url = `${this.url}skus.csv`;
+    options = Object.assign(options, {
+        url: url,
+        method: "GET",
+        auth: true
+    });
+    options = this._build(options);
+    console.log("OPTIONS", options);
```
joao-conde, line 82: console log

joao-conde (changes requested):

joao-conde (approved):

src/js/api/sku.js
```diff
@@ -280,3 +323,46 @@ ripe.Ripe.prototype.validateSkuP = function(id, options) {
...
     });
 };
+
+/**
+ * Counts the existing SKUs, according to the provided filtering
+ * strategy as normalized values.
+ *
+ * @param {Object} options An object of options to configure the request, such as:
+ * - 'filters[]' - List of filters that the query will use to, operators such as
+ * ('in', 'not_in', 'like', 'contains'), for instance (eg: 'id:eq:42') would filter by the id that equals to 42.
```
joamag, line 333: I don't think we need to be explicit about the way the filters are defined... otherwise, these comments are going to be very difficult to maintain

src/js/api/sku.js
```diff
@@ -54,6 +54,49 @@ ripe.Ripe.prototype.getSkusP = function(options) {
     });
 };
 
+/**
+ * Gets the existing SKUs in a CSV file, according to the provided filtering
+ * strategy as normalized values.
+ *
+ * @param {Object} options An object of options to configure the request, such as:
+ * - 'filters[]' - List of filters that the query will use to, operators such as
```
joamag, line 62: same here

joamag (changes requested):

src/js/api/sku.js
BeeMargarida: Got it, but this is also in a lot of other places, do we also remove it there?

joamag (approved):

PR #371, opened 2022-03-11
fix: optional brand in payload

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/281 |
| Dependencies | -- |
| Decisions | Brand is optional so should not be used in the payload by default. |

### DIFF
--- a/src/js/api/bulk-order.js
+++ b/src/js/api/bulk-order.js
@@ -535,7 +535,6 @@ ripe.Ripe.prototype._importBulkOrder = function(name, orders, options = {}) {
     const description = options.description === undefined ? null : options.description;
     const dataJ = {
         name: name,
-        brand: brand,
         orders: orders
     };
     if (brand) dataJ.brand = brand;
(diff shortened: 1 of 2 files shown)


src/js/api/bulk-order.js
```diff
@@ -535,7 +535,6 @@ ripe.Ripe.prototype._importBulkOrder = function(name, orders, options = {}) {
     const description = options.description === undefined ? null : options.description;
     const dataJ = {
         name: name,
-        brand: brand,
```
BeeMargarida, line 538: isn't it already optional? Since `const brand` can be null

joao-conde: But it is preferably to leave it as the line below:

```js
if (brand) dataJ.brand = brand;
```

BeeMargarida: oops, I didn't see it :eyes: you are right

BeeMargarida (approved):

src/js/api/bulk-order.js
joao-conde: I was the one making the change below :p But I missed this.

3rdvision (approved):

joamag (approved):

PR #373, opened 2022-03-15
#4703 Add size to rules

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-core/issues/4703 |
| Dependencies | https://github.com/ripe-tech/ripe-core/pull/4707 |
| Decisions | - Adapt ripe-sdk methods to new size dimensions in rules. <br> - Add missing resolveTransportRule and resolveTransportRuleP methods.  |

### DIFF
--- a/src/js/api/transport-rule.js
+++ b/src/js/api/transport-rule.js
@@ -246,7 +246,7 @@ ripe.Ripe.prototype.resolveTransportRule = function(
 };
 
 /**
- * Gets an existing transport rule filtered by brand, model and country.
+ * Gets an existing transport rule filtered by brand, model, country, factory and size.
  *
  * @param {String} brand The transport rule's brand.
  * @param {String} model The transport rule's model.
--- a/src/js/api/transport-rule.js
+++ b/src/js/api/transport-rule.js
@@ -194,3 +194,87 @@ ripe.Ripe.prototype.deleteTransportRuleP = function(id, options) {
         });
     });
 };
+
+/**
+ * Gets an existing transport rule filtered by brand, model and country.
+ *
+ * @param {String} brand The transport rule's brand.
+ * @param {String} model The transport rule's model.
+ * @param {String} country The transport rule's country.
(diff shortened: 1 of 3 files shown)


BeeMargarida (changes requested):
Shouldn't this be done for `transport-rule` and `availability-rule` as well?

3rdvision:
> Shouldn't this be done for `transport-rule`, `availability-rule` and `price-rule` as well?

for which methods? Everything is passed through options and as for the resolve method, there's only `resolveInvoiceRule`

BeeMargarida:
> > Shouldn't this be done for `transport-rule`, `availability-rule` and `price-rule` as well?
> 
> for which methods? Everything is passed through options and as for the resolve method, there's only `resolveInvoiceRule`

Nvm, i did not notice those don't have resolve methods. However, `transport-rule` should have (I guess it wasn't ported at the time) because there is an endpoint for it - https://github.com/ripe-tech/ripe-core/blob/7a0a2eed80467916980ecd02f487d307ca5cfa60/src/ripe_core/controllers/api/rules/transport_rule.py#L49
Perhaps this PR should add it, since it is basically the same as the one for invoice rule

src/js/api/transport-rule.js
```diff
@@ -194,3 +194,87 @@ ripe.Ripe.prototype.deleteTransportRuleP = function(id, options) {
         });
     });
 };
+
+/**
+ * Gets an existing transport rule filtered by brand, model and country.
```
BeeMargarida, line 199:
```suggestion
 * Gets an existing transport rule filtered by brand, model, country, factory and size.

```

BeeMargarida (approved):

joao-conde (approved):

3rdvision:
blocked until https://github.com/ripe-tech/ripe-core/pull/4707 is merged

joamag (approved):

PR #374, opened 2022-03-22
#317: filter and fetch justifications by code

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/317 |
| Dependencies | -- |
| Decisions | Moving the default logic (of resolving `context:code` or just `code_full`) to the SDK. Also hides away the logic of the `other;` "escape hatch". This logic would have to be repeated in clients like RIPE Robin or RIPE Pulse (and future ones). |

image

image

### DIFF
--- a/src/js/api/justification.js
+++ b/src/js/api/justification.js
@@ -24,7 +24,6 @@ if (
  * - 'skip' - The number of the first record to retrieve from the results.
  * - 'limit' - The number of results to retrieve.
  * @param {Function} callback Function with the result of the request.
- * @param {Function} callback Function with the result of the request.
  * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
  */
 ripe.Ripe.prototype.getJustifications = function(options, callback) {
@@ -107,6 +106,21 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
     });
 };
 
+/**
+ * Gets the existing justifications, according to the provided filtering
+ * strategy as normalized values.
+ *
+ * Attempts to resolve context, code and full code based on what is given.
+ * Adds those parameters to the filters given.
+ *
+ * If the 'other' code is used, it returns the adhoc text message without
+ * requesting a justification from RIPE Core.
+ *
+ * @param {Object} options An object of options to configure the request.
+ * @param {Function} callback Function with the result of the request.
+ * @param {String} other The 'other' prefix code.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
 ripe.Ripe.prototype._getJustifications = function(options, callback, other = "other;") {
     // resolve context, code and full code based on available
     // information, using smarter filters
@@ -135,6 +149,13 @@ ripe.Ripe.prototype._getJustifications = function(options, callback, other = "ot
     return this._cacheURL(options.url, options, callback);
 };
 
+/**
+ * Attempts to resolve context, code and full code based on what is given.
+ * A full code is equivalent to 'context:code'.
+ *
+ * @param {Object} options An object of options that may include context, code or a full code.
+ * @returns {Object} The params to be added as filters like context, code and full code.
+ */
 ripe.Ripe.prototype._resolveJustificationParams = function(options) {
     const params = {};
     if (options.codeFull) {
--- a/src/js/api/justification.js
+++ b/src/js/api/justification.js
@@ -37,7 +37,7 @@ ripe.Ripe.prototype.getJustifications = function(options, callback) {
         auth: true
     });
     options = this._build(options);
-    return this._cacheURL(options.url, options, callback);
+    return this._getJustifications(options, callback);
 };
 
 /**
@@ -59,6 +59,7 @@ ripe.Ripe.prototype.getJustificationsP = function(options) {
  * Gets the existing justifications filtered by context, according to the
  * provided filtering strategy as normalized values.
  *
+ * @param {String} context The justification context to filter for.
  * @param {Object} options An object of options to configure the request, such as:
  * - 'filters[]' - List of filters that the query will use to, operators such as
  * ('in', 'not_in', 'like', 'contains'), for instance (eg: 'id:eq:42') would filter by the id that equals to 42.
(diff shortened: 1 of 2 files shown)


src/js/api/justification.js
```diff
@@ -104,3 +105,69 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+
+/**
+ * Gets the existing justifications, according to the provided filtering
+ * strategy as normalized values.
+ *
+ * Attempts to resolve context, code and full code based on what is given.
+ * Adds those parameters to the filters given.
+ *
+ * If the 'other' code is used, it returns the adhoc text message without
+ * requesting a justification from RIPE Core.
```
BeeMargarida, line 117: Perhaps add explanation on why the other logic is different, like (this is just a suggestion). Perhaps also add that explanation in the comment added above the code

```suggestion
 * Gets the existing justifications, according to the provided filtering
 * strategy as normalized values.
 *
 * Attempts to resolve context, code and full code and adds those
 * parameters to the filters provided.
 *
 * If the 'other' keyword is contained in the provided `codeFull`, it means
 * that no default justification was used and the custom text provided is
 * the text that is returned.

```

BeeMargarida (commented):
I think that this is clean 👌

src/js/api/justification.js
joao-conde: I like it

BeeMargarida (approved):

3rdvision (approved):

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+            context: "other",
+            code_full: params.codeFull,
+            text: params.codeFull.slice(index + other.length)
+        };
+        return callback([justification], true, null);
+    }
+
+    // builds the appropriate filters for context, code
+    // and full code, used to fetch the justification
+    options.params = options.params !== undefined ? options.params : {};
```
joamag, line 144: use positive comparison

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+};
+
+/**
+ * Attempts to resolve context, code and full code based on what is given.
+ * A full code is equivalent to 'context:code'.
+ *
+ * @param {Object} options An object of options that may include context, code or a full code.
+ * @returns {Object} The params to be added as filters like context, code and full code.
+ */
+ripe.Ripe.prototype._resolveJustificationParams = function(options) {
```
joamag, line 160: build instead of resolve

joamag: maybe options for the name instead of params

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+        };
+        return callback([justification], true, null);
+    }
+
+    // builds the appropriate filters for context, code
+    // and full code, used to fetch the justification
+    options.params = options.params !== undefined ? options.params : {};
+    options.params.filters = options.params.filters !== undefined ? options.params.filters : [];
+    if (params.context) options.params.filters.push(`context:likei:${params.context}`);
+    if (params.code) options.params.filters.push(`code:likei:${params.code}`);
```
joamag, line 147: equals

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+ * If the 'other' keyword is contained in the provided `codeFull`, it means
+ * that no RIPE Core justification was used and the custom text provided is
+ * the text that is returned.
+ *
+ * @param {Object} options An object of options to configure the request.
+ * @param {Function} callback Function with the result of the request.
+ * @param {String} other The 'other' prefix code.
+ * @returns {XMLHttpRequest} The XMLHttpRequest instance of the API request.
+ */
+ripe.Ripe.prototype._getJustifications = function(options, callback, other = "other;") {
```
joamag, line 125: resolveJustification

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+    // resolve context, code and full code based on available
+    // information, using smarter filters
+    const params = this._resolveJustificationParams(options);
```
joamag, line 128: early exit when no $ is present

src/js/api/justification.js
```diff
@@ -104,3 +105,70 @@ ripe.Ripe.prototype.getJustificationsByContextP = function(context, options) {
...
+
+    // exits if the 'other' context is used which works
+    // as an escape hatch and does not need to be looked up
+    if (params.codeFull && params.codeFull.includes(other)) {
```
joamag, line 132: startsWith

src/js/api/justification.js
joamag, line 125: separator = ";"

joamag (changes requested):

joao-conde:
@joamag just an idea but maybe this could be moved to an endpoint on the RIPE Core justifications API? Because otherwise all this logic will be repeated across API clients (JS, python, android and iOS SDKs)

joamag:
> @joamag just an idea but maybe this could be moved to an endpoint on the RIPE Core justifications API? Because otherwise all this logic will be repeated across API clients (JS, python, android and iOS SDKs)

Agree let's move this to the server-side

src/js/api/justification.js
joao-conde: After some thinking, I find better if the $ handling remains in the client. Otherwise, will the $ be in the code full, or in the context, when one is given? That does not map well for developers using the RIPE Core API.

In other words, this intuitively seems better (because it is a one-on-one mapping of variables to model fields):
image


Than this:
image

joao-conde:
Closing due to favor of https://github.com/ripe-tech/ripe-core/pull/4721

PR #378, opened 2022-03-24
YSL: Add `remoteInitialsBuilderLogic` option

3rdvision:
| - | - |
| --- | --- |
| Decisions | - Add `remoteInitialsBuilderLogic` and `noAwaitLayout` options. <br> - Adapt to not use initials builder in the client side and the cancellation logic to allow for faster response times. |

### DIFF
--- a/src/js/base/ripe.js
+++ b/src/js/base/ripe.js
@@ -683,16 +683,13 @@ ripe.Ripe.prototype.setOptions = function(options = {}) {
     this.usePrice = this.options.usePrice === undefined ? !this.noPrice : this.options.usePrice;
     this.noDiag = this.options.noDiag === undefined ? false : this.options.noDiag;
     this.useDiag = this.options.useDiag === undefined ? !this.noDiag : this.options.useDiag;
+    this.noAwaitLayout =
+        this.options.noAwaitLayout === undefined ? false : this.options.noAwaitLayout;
     this.useInitialsBuilderLogic =
         this.options.useInitialsBuilderLogic === undefined
             ? true
             : this.options.useInitialsBuilderLogic;
-    this.remoteInitialsBuilderLogic =
-        this.options.remoteInitialsBuilderLogic === undefined
-            ? false
-            : this.options.remoteInitialsBuilderLogic;
-    this.noAwaitLayout =
-        this.options.noAwaitLayout === undefined ? false : this.options.noAwaitLayout;
+    this.composeLogic = this.options.composeLogic === undefined ? false : this.options.composeLogic;
 
     // in case the requested format is the "dynamic" lossless one
     // tries to find the best lossless image format taking into account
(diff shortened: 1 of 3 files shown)


joamag:
@3rdvision adding fast-track as it's important to move this to master as soon as possible ;)

src/js/base/ripe.js
```diff
@@ -1380,7 +1384,7 @@ ripe.Ripe.prototype.update = async function(state = null, options = {}, children
 
     try {
         this.updatePromise = _update();
-        if (options.noAwaitLayout) return true;
+        if (options.noAwaitLayout || this.remoteInitialsBuilderLogic) return true;
```
BeeMargarida, line 1387: Why this? Shouldn't the flag just control if the `initialsBuilder` is called and the `logic` param in the compose URL?

3rdvision: Unfortunately it wasn't that simple. The problem is: by default we have an await for the last update logic which means that if you do 2 requests one after the other (example "1" and after "12") the second one will await the first one. We don't want that behavior when we `remoteInitialsBuilderLogic` is set to true as it would create that problem of rendering the "1" and then only the "12" as the cancelling logic is in the initials builder promise which is not used here.

BeeMargarida: This this update is also can include the configurator. Won't this break any behaviour with that?

3rdvision: Note that this doesn't alter the default behavior of the SDK - it's not a breaking change.
Only if you set the option `remoteInitialsBuilderLogic` to true will the SDK behavior differently and not wait the first update to be complete before processing the second one.

Also, I'm not sure how it would break but anyway I tested this customizing a swear vyner with the `remoteInitialsBuilderLogic` set to true and everything works. In fact, I found it to be even faster when changing parts but are you thinking a specific scenario?

BeeMargarida: What I'm saying is that this is a specific logic for image, so it should not imply changes for the overall behaviour of SDK but only in its image. For example, it makes sense to not await when the variable `noAwaitLayout` is set (it's right in the name), but it is weird that the configuration for remote iniitals builder changes that logic (when it is only used in image)
One possible problem might appear using the preferred frame logic in the configurator (can be enabled in the White technical modal) and switching parts fast.

BeeMargarida: @3rdvision I know this is p-high, so if you want you can move this forward and then you talk better about this with João

3rdvision: I added `noAwaitLayout` as an option.
In that scenario, by default the SDK would have `noAwaitLayout` set to false and that way it wouldn't happen.

BeeMargarida (approved):
I'm still not entirely convinced but I'll approve it since it is fast track :runner:

joao-conde (approved):

joamag (approved):

PR #393, opened 2022-05-13
#135 Add `fromArrayBuffer` and `fromBlob` to `FileTuple`

NFSS10:
| - | - |
| --- | --- |
| Issue | Port from https://github.com/hivesolutions/yonius/pull/39 (related to https://github.com/ripe-tech/hermes-proxy/issues/135) |
| Dependencies | -- |
| Decisions | • Add `fromArrayBuffer` and `fromBlob` to `FileTuple`<br>• Add tests: `#fromString()`, `#fromArrayBuffer()` and `#fromBlob()` |

imagem

imagem

### DIFF
--- a/test/js/base/struct.js
+++ b/test/js/base/struct.js
@@ -38,6 +38,12 @@ describe("FileTuple", function() {
     });
 
     describe("#fromString()", async function() {
+        beforeEach(function() {
+            if (typeof TextEncoder === "undefined") {
+                this.skip();
+            }
+        });
+
         it("should be able to create a simple file tuple objects", () => {
             const fileTuple = ripe.ripe.FileTuple.fromString("hello", "hello.txt", "text/plain");
             assert.notStrictEqual(fileTuple, null);
@@ -51,6 +57,12 @@ describe("FileTuple", function() {
     });
 
     describe("#fromArrayBuffer()", async function() {
+        beforeEach(function() {
+            if (typeof TextEncoder === "undefined") {
+                this.skip();
+            }
+        });
+
         it("should be able to create a simple file tuple objects", () => {
             const fileTuple = ripe.ripe.FileTuple.fromArrayBuffer(
                 new ArrayBuffer(),
@@ -69,7 +81,7 @@ describe("FileTuple", function() {
 
     describe("#fromBlob()", async function() {
         beforeEach(function() {
-            if (typeof Blob === "undefined") {
+            if (typeof TextEncoder === "undefined" || typeof Blob === "undefined") {
                 this.skip();
             }
         });
--- a/test/js/base/struct.js
+++ b/test/js/base/struct.js
@@ -57,12 +57,6 @@ describe("FileTuple", function() {
     });
 
     describe("#fromArrayBuffer()", async function() {
-        beforeEach(function() {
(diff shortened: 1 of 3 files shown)


test/js/base/struct.js
```diff
@@ -34,4 +36,69 @@ describe("FileTuple", function() {
...
+            assert.strictEqual(fileTuple instanceof Array, true);
+        });
+    });
+
+    describe("#fromArrayBuffer()", async function() {
+        beforeEach(function() {
+            if (typeof TextEncoder === "undefined") {
+                this.skip();
+            }
+        });
```
joao-conde, line 64: This one does not need the TextEncoder right?

test/js/base/struct.js
```diff
@@ -34,4 +36,69 @@ describe("FileTuple", function() {
...
+            assert.strictEqual(fileTuple instanceof Array, true);
+        });
+    });
+
+    describe("#fromBlob()", async function() {
+        beforeEach(function() {
+            if (typeof TextEncoder === "undefined" || typeof Blob === "undefined") {
+                this.skip();
+            }
+        });
```
joao-conde, line 87: Same for this one, does not need TextEncoder right?

test/js/base/struct.js
```diff
@@ -2,6 +2,8 @@ const assert = require("assert");
 const config = require("../config");
 const ripe = require("../../../src/js");
 
+const { Blob } = require("buffer");
```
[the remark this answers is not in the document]
joao-conde: If there are collisions we can rename the variables to _buffer or buffer_ or if the number of places to edit are too many we can import this as `bufferNode`

joao-conde (changes requested):
Looks good for the most part, I would also order the fromX methods by alphabetical order (and corresponding test suites)

test/js/base/struct.js
NFSS10: Nope, this needs to stay.

`fromArrayBuffer` calls `fromData`, which then call `encodeMultipart` which uses `TextEncoder` and it breaks if that doesn't exist.

test/js/base/struct.js
NFSS10: Nope, this needs to stay.

`fromBlob` calls `fromArrayBuffer` -> `fromData` -> `encodeMultipart` which uses `TextEncoder` and it breaks if that doesn't exist.

NFSS10:
> Looks good for the most part, I would also order the fromX methods by alphabetical order (and corresponding test suites)

As to the order, it takes into the account the methods declaration as is more important to follow than alphabetical order (makes the code easier to read)

`fromData` -> independent
`fromString` -> uses `fromData`
`fromArrayBuffer` -> uses `fromData`
`fromBlob` -> uses `fromArrayBuffer`

test/js/base/struct.js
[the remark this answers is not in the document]
joamag: yes, we have some exceptions (eg: tobias-bot and slack bot mostly because the examples of equivalent bots use this standard) but when using common js lets import the complete object instead of unpacking its values

joamag (changes requested):

joamag (approved):

PR #394, opened 2022-05-18
#978: support profile overriding in image

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/978 |
| Dependencies | -- |
| Decisions | - Support for profile overriding on image - it should be possible to override the profiles, this is useful for showing personalization frames that are defined by thumbnails - see https://github.com/ripe-tech/ripe-docs/pull/618 |
| Animated GIF | -- |

### DIFF
--- a/src/js/visual/image.js
+++ b/src/js/visual/image.js
@@ -106,6 +106,7 @@ ripe.Image.prototype.init = function() {
     this.initialsGroup = this.options.initialsGroup || null;
     this.initialsContext = this.options.initialsContext || null;
     this.getInitialsContext = this.options.getInitialsContext || null;
+    this.initialsProfiles = this.options.initialsProfiles || null;
     this.initialsBuilder = this.options.initialsBuilder || this.owner.initialsBuilder;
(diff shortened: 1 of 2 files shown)


src/js/visual/image.js
```diff
@@ -436,6 +440,10 @@ ripe.Image.prototype.update = async function(state, options = {}) {
             return false;
         }
 
+        const profile = initialsProfiles
+            ? [...initialsProfiles, ...initialsSpec.profile]
```
3rdvision, line 444: shouldn't this be the inverse to allow more freedom when overriding?
```suggestion
            ? [...initialsSpec.profile, ...initialsProfiles]

```

BeeMargarida: No, since the profiles we provide in the spec should take precendence, or else it won't show in some cases

3rdvision (approved):

joao-conde (approved):

joamag (approved):

PR #402, opened 2022-06-29
fix: import order DKU

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Contents filled only if there are values. If there are none, dont send an object with null values. Pass DKU if it exists. Add a size to import order.<br> Passing no size, previously, we had a contents like `{ size: null }` and this does not fail when importing. Meaning there is some bug in Core, to be fixed in another PR. |
| Animated GIF | -- |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Add dku to request options for import order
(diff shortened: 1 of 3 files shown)


joao-conde:
CI failing due to SBX probably an example of flaky tests, this passes local tests and the order can be imported correctly even against SBX

3rdvision (approved):

joamag (approved):

PR #413, opened 2022-09-05
Set CSR camera standard based on a cube

NFSS10:
| - | - |
| --- | --- |
| Issue | The camera default values should be set to support every build by default |
| Dependencies | -- |
| Decisions | - Changed ripe-sdk demo brand and model to `dummy` `cube` as it's the build with the best CSR support<br>- Update the camera default values so it uses standard values |

https://user-images.githubusercontent.com/22588915/188492870-15dc52e6-e8eb-4443-b34b-7e90b6bbe8f0.mp4


NFSS10:
Closing as the standard may still change

PR #427, opened 2022-09-30
#429 Loading CSR based on spec

NFSS10:
| - | - |
| --- | --- |
| Dependencies | - <del>https://github.com/ripe-tech/ripe-sdk/pull/431</del><br/>- https://github.com/ripe-tech/ripe-docs/pull/625 |
| Decisions | - Move core CSR initialization logic to on `post_config` event instead of just on the first `init`<br>- Update which values should be updated by the configurator and which values should be updated by the config<br>- Allow to choose between asynchronously and synchronously setup |

https://user-images.githubusercontent.com/22588915/207101956-b9c1dcf8-1c38-4dfb-b8d8-46b198478aaa.mp4

**Test config used for testing this feature (just adds it to the start of `_onPostConfig()`:**
```js
const specInitials = {
            "3d": {
                ...(config.initials["3d"] || {}),
                width: 3000,
                height: 300,
                font_size: 280,
                stoke_width: 5,
                text_displacement_blur: 1.5,
                text_normal_map_blur: 1,
                points: [
                    [-1, 0, 0],
                    [0, 0, -1],
                    [1, 0, 0]
                ],
                position: [0, 0, 0],
                rotation: [0, 0, 0],
                scale: 1,
                material_color: "ffffff",
                material_metalness: 0,
                material_roughness: 1,
                material_emissive_color: "000000",
                material_emissive_intensity: 1,
                material_displacement_scale: 25,
                material_displacement_bias: 0,
                mesh_width_segments: 1000,
                mesh_height_segments: 100
            }
        };
        const spec3d = {
            scene: {
                environment: "studio_small_2",
                tone_mapping: "aces_filmic",
                tone_mapping_exposure: 0.8,
                camera: {
                    position: [0, 0, 207],
                    rotation: [0, 0, 0],
                    fov: 24.678,
                    film_gauge: 35,
                    aspect: 1,
                    near: 0.1,
                    far: 10000
                },
                camera_look_at: [0, 0, 0],
                zoom: {
                    enabled: true,
                    min: 0.75,
                    max: 1.5,
                    sensitivity: 1
                }
            }
        };
        config.initials = { ...config.initials, ...specInitials };
        config["3d"] = spec3d;
```


src/js/visual/configurator-csr.js
```diff
@@ -1333,6 +1354,166 @@ ripe.ConfiguratorCsr.prototype._onInitialsExtraEvent = function(self, initialsEx
...
+
+        // renders newly build scene
+        this._render();
+
+        self.loading = false;
+        this.trigger("ready");
+    };
+
+    // runs synchronously or asynchronously depending on how the CSR configurator was setup
+    this.awaitPostConfig ? await _postConfig() : _postConfig();
```
3rdvision, line 1514: what difference does this make for the user?

src/js/visual/configurator-csr.js
```diff
@@ -1363,3 +1544,19 @@ ripe.ConfiguratorCsr.prototype._unregisterInitialsHandlers = function() {
     this.owner && this.owner.unbind("initials_extra", this._onInitialsExtraEvent);
     this.owner && this.owner.unbind("initials", this._onInitialsEvent);
 };
+
+/**
+ * @ignore
+ */
+ripe.ConfiguratorCsr.prototype._registerConfigHandlers = function() {
+    this.owner.bind("pre_config", () => this._onPreConfig(this));
```
3rdvision, line 1552: can we keep this consistent like this, even if we're not using?
```suggestion
    this.owner.bind("pre_config", config => this._onPreConfig(this, config));

```

src/js/visual/configurator-csr.js
```diff
@@ -897,14 +870,19 @@ ripe.ConfiguratorCsr.prototype._initCsrRenderedInitials = async function() {
...
  */
 ripe.ConfiguratorCsr.prototype._deinitCsrRenderedInitials = function() {
     // cleanup handlers
     this._unregisterInitialsHandlers();
+};
 
+/**
+ * Free resources used by the csr initials.
+ */
+ripe.ConfiguratorCsr.prototype._unloadCsrRenderedInitialsResources = function() {
```
3rdvision, line 885: I'd simplify to just this... what do you think?
I'd do this for the others with Resources sufix
```suggestion
ripe.ConfiguratorCsr.prototype._unloadCsrRenderedInitials = function() {

```

src/js/visual/configurator-csr.js
```diff
@@ -184,10 +139,9 @@ ripe.ConfiguratorCsr.prototype.init = function() {
     // creates the necessary DOM elements and runs the
     // CSR initializer
     this._initLayout();
-    this._initCsr().then(async () => {
-        this.loading = false;
-        await this.flushPending(true);
```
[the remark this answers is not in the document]
NFSS10: Didn't make sense anymore

src/js/visual/configurator-csr.js
NFSS10: It allows the loading to by asynchrounsly thus allowing blocking the thread

src/js/visual/configurator-csr.js
NFSS10: Yes!
Altough it would be `brand, model, options` 👍

src/js/visual/configurator-csr.js
NFSS10: If you check the whole deinit logic it has like three methods. Each with a specific tas Ex for scene:

`_deinitScene()` - Destroys scene without disposing it's resources
`_unloadSceneResources()` - Disposes resources related to the scene, as in textures, meshes, materials, ... (it's basically cleanup of resources that the GC doesn't handle).
`_destroyScene`- Uses the other two methods. It completly destroys the scene and everything related to it.

So, if we want to change the names, I would keep something that indicates that there is a difference between them, that's why I added the "resources" suffix

src/js/visual/configurator-csr.js
3rdvision: nevermind then!

3rdvision (approved):

joamag:
@NFSS10 let's schedule a meeting to start the discussion of the spec adaption

joao-conde (approved):

Issue #429, opened 2022-10-03
Client-Side Rendering (CSR)


joao-conde:
## Sponsors

- João Magalhães - <joao@platforme.com>
- João Conde - <jc@platforme.com>
- Nuno Silva - <ns@platforme.com>

## Rationale

Pre-Render & Compose (PRC) offers maximum quality but low interactivity for customers. Client-Side Rendering allows for maximum interactivity at the cost of some quality. Additionally, CSR is critical in order to provide state-of-the-art rendering strategies.

We're aiming at using this new CSR approach as a tool to strengthen our relationship with our biggest client (Dior).

## Description

Extend RIPE SDK to provide a real 3D configurator making use of WebGL technology and using our 3DBs as data source.

### Software development

- Define a spec for the 3D information to be stored in the 3DB
- Extend RIPE SDK with a new configurator
- Implement the current SDK API, and extend if need be

## Estimation

| Discipline | Estimation |
| --- | --- |
| **Software Development** | 60 days |

NFSS10:
@hugo-gomes This issue is paused because the PRs are opened for some weeks now. If you take some time to do the final approval and merge them that would be great!

hugo-gomes:
@NFSS10 let's schedule a meeting to discuss those https://calendly.com/d/xz5-ngz-378/client-side-rendering-csr-429

joao-conde:
Believe this can be closed as CSR is not going to be done for RIPE SDK and instead a bespoke solution for Realtime.

PR #431, opened 2022-10-04
Fix CSR debug mode not properly initializing when no CSR initials points are set

NFSS10:
| - | - |
| --- | --- |
| Dependencies | -- |
| Decisions | Fixed issue where CSR debug tools wouldn't initialize if the CSR initials were enabled and the CSR initials points array had no points |

### DIFF
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -1025,34 +1025,37 @@ ripe.ConfiguratorCsr.prototype._initDebug = function() {
             this.debugRefs.renderedInitials.group.add(this.debugRefs.renderedInitials.axis);
         }
 
-        // inits reference points curve
-        if (this.debugOptions.renderedInitials.line) {
-            const curve = new window.THREE.CatmullRomCurve3(
-                this.initialsRefs.renderedInitials.points,
-                false,
-                "centripetal"
-            );
-            const pointsNum = 50;
-            const linePoints = curve.getPoints(pointsNum);
-            this.debugRefs.renderedInitials.line = new window.THREE.Line(
-                new window.THREE.BufferGeometry().setFromPoints(linePoints),
-                new window.THREE.LineBasicMaterial({ color: 0x0000ff })
-            );
+        // ensures it has the minimum number of points
+        if (this.initialsRefs.renderedInitials.points > 1) {
+            // inits reference points curve
+            if (this.debugOptions.renderedInitials.line) {
+                const curve = new window.THREE.CatmullRomCurve3(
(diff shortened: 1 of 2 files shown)


src/js/visual/configurator-csr.js
```diff
@@ -1025,34 +1025,37 @@ ripe.ConfiguratorCsr.prototype._initDebug = function() {
...
-            const linePoints = curve.getPoints(pointsNum);
-            this.debugRefs.renderedInitials.line = new window.THREE.Line(
-                new window.THREE.BufferGeometry().setFromPoints(linePoints),
-                new window.THREE.LineBasicMaterial({ color: 0x0000ff })
-            );
-
-            this.debugRefs.renderedInitials.group.add(this.debugRefs.renderedInitials.line);
-        }
+        // ensures it has the minimum number of points
+        if (this.initialsRefs.renderedInitials.points > 1) {
```
NFSS10, line 1029: **For the reviewers:**
The fix is very small, it was just adding this check, the rest of the changes is just the code now being inside this if brackets

joao-conde, line 1029: Do you mean to do "points.length" ? Or points is a counter?

NFSS10: Ups, made a mistake when moving the fix to this branch 👀

src/js/visual/configurator-csr.js
```diff
@@ -1025,34 +1025,37 @@ ripe.ConfiguratorCsr.prototype._initDebug = function() {
...
-            const linePoints = curve.getPoints(pointsNum);
-            this.debugRefs.renderedInitials.line = new window.THREE.Line(
-                new window.THREE.BufferGeometry().setFromPoints(linePoints),
-                new window.THREE.LineBasicMaterial({ color: 0x0000ff })
-            );
-
-            this.debugRefs.renderedInitials.group.add(this.debugRefs.renderedInitials.line);
-        }
+        // ensures it has the minimum number of points
+        if (this.initialsRefs.renderedInitials.points.length > 1) {
```
joao-conde, line 1029: Ok, sorry again, but isnt the minimum 1? In other words, dont you want this to be > 0?
If it is some ad-hoc maximum value should that be hardcoded or softcoded as an argument?

NFSS10: No, I told you the min and max can vary.

NFSS10: No, I wan't the minimum to be 2 as you can only have a line with a minimum of 2 points.

PR #445, opened 2022-11-22
fix: no mask buffers if useMasks false

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/446 |
| Dependencies | -- |
| Decisions | If masks are turned off, we cant create the mask buffers (image elements) because they will load the masks anyways even if disabled. |
| Animated GIF | -- |

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -399,30 +399,35 @@ ripe.ConfiguratorPrc.prototype.resize = async function(size, width, height) {
     }
 
     const area = this.element.querySelector(".area");
-    const frontMask = this.element.querySelector(".front-mask");
     const back = this.element.querySelector(".back");
-    const mask = this.element.querySelector(".mask");
     area.width = width * this.pixelRatio;
     area.height = height * this.pixelRatio;
     area.style.width = width + "px";
     area.style.height = height + "px";
-    frontMask.width = width;
-    frontMask.height = height;
-    frontMask.style.width = width + "px";
-    frontMask.style.height = height + "px";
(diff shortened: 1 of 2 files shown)


src/js/visual/configurator-prc.js
```diff
@@ -399,30 +399,35 @@ ripe.ConfiguratorPrc.prototype.resize = async function(size, width, height) {
...
     back.style.marginLeft = `-${String(width)}px`;
-    mask.width = width;
-    mask.height = height;
-    mask.style.width = width + "px";
-    mask.style.height = height + "px";
     this.currentSize = size;
     this.currentWidth = width;
     this.currentHeight = height;
+
+    if (this.useMasks) {
```
3rdvision, line 416: I'd say that this should also be passing if `this.useMasks` is null or even undefined

```suggestion
    if (this.useMasks !== false) {

```

src/js/visual/configurator-prc.js
```diff
@@ -1018,15 +1018,21 @@ ripe.ConfiguratorPrc.prototype._initLayout = function() {
     // temporarily store the images of the product's frames
     const framesBuffer = ripe.createElement("div", "frames-buffer");
 
-    // creates a masksBuffer element that will be used to store the continuous
-    // mask images to be used during highlight and select operation
-    const masksBuffer = ripe.createElement("div", "masks-buffer");
+    if (this.useMasks) {
```
3rdvision, line 1021: same for other `if (this.useMasks)`

3rdvision (changes requested):

joao-conde:
Closing this as the solution is not ideal

Issue #446, opened 2022-11-22
No masks options disable highlighting but masks are still loaded


joao-conde:
## Description

Through the use of `noMasks` or `useMasks` one can control whether the masks are used. However, even when not using them, the SDK still loads them and these represent a significant amount of network requests. We should not load masks if they are not going to be used.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The masks are show and not loaded i.e. no network requests for masks are made. |
| **Observed** | The masks are not shown but are loaded and you can see it in the network requests. |

## Repro Steps

1. Initialize our RIPE in this repo
2. Set `useMasks: false` or `noMasks: true` in the `bindConfigurator` method
3. Build the SDK, launch the demo, go to localhost:8080/simple network tab

## Screens

![image](https://user-images.githubusercontent.com/16060539/203329650-c1841c45-78d2-426c-b2a3-2b07d809e32d.png)

msp-platforme:
:+1: 
When masks are on (parameter use_masks=1) we see a lot of network requests and we can also hover the model and see the masks.
![image](https://user-images.githubusercontent.com/72499843/204344707-79161e03-29f0-488f-ace8-d60e8a1a1d5d.png)

When masks are off (parameter use_masks=0) we don't see network requests related to maks and we if we hover the model it does not have masks.
![image](https://user-images.githubusercontent.com/72499843/204344907-baf932ab-e532-448e-95a6-ae0ee01c2459.png)

PR #447, opened 2022-11-23
fix: load mask only if using masks

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/446 |
| Dependencies | -- |
| Decisions | If masks are not to be used, just avoid loading them but create the buffers anyways. |
| Animated GIF | **Before**<br> image **After**<br>  image|

For example, using the following test script:

```js
configuratorPrc = ripe.bindConfigurator(element, {
    duration: 250,
    useMasks: true,
    view: bestFace(result)
});
configuratorPrc.isFirst = true;

setTimeout(() => {
    configuratorPrc.updateOptions({ useMasks: false });
}, 6000);
```

I was able to start out with masks and 6 seconds later have them disappear and requests stopped being made:

disablemasks

### DIFF
--- a/src/js/visual/configurator-prc.js
+++ b/src/js/visual/configurator-prc.js
@@ -1240,7 +1240,7 @@ ripe.ConfiguratorPrc.prototype._loadFrame = async function(view, position, optio
 
     // triggers the async loading of the "master" mask for the current
     // frame, this should imply some level of cache usage
-    this._loadMask(maskImage, view, position, options);
+    if (this.useMasks) this._loadMask(maskImage, view, position, options);
 
     // apply pixel ratio to image dimensions so that the image obtained
     // reflects the target pixel density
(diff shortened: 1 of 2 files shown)


src/js/visual/configurator-prc.js
```diff
@@ -1240,7 +1240,7 @@ ripe.ConfiguratorPrc.prototype._loadFrame = async function(view, position, optio
 
     // triggers the async loading of the "master" mask for the current
     // frame, this should imply some level of cache usage
-    this._loadMask(maskImage, view, position, options);
+    if (this.useMasks) this._loadMask(maskImage, view, position, options);
```
hugo-gomes, line 1243: add this clause logic to the comment above please

hugo-gomes (changes requested):

hugo-gomes (approved):

PR #448, opened 2022-12-06
#271 Add quantity param to import order

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/271 |
| Decisions | - Add `quantity` param to `_importOrder` method |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* Add `quantity` param to `_importOrder` method - [ripe-pulse/#271](https://github.com/ripe-tech/ripe-pulse/issues/271)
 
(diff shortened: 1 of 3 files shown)


3rdvision:
I was able to run these tests against ripe-core master and I believe we should ignore it failing for now.

The tests are not passing for 2 reasons:

1- RIPE Core SANDBOX downtime is bad which makes it hard to run tests as the probability that the server is down because it's installing some build is high.

2- I'm not sure if `config.TEST_URL=https://ripe-core-sbx.platforme.com/api/` or if it's another environment (I don't have access) and that ripe-core might not be the latest version - it might be without this patch https://github.com/ripe-tech/ripe-core/pull/4766

joao-conde:
@3rdvision I think for tests the default is being used, which is sandbox.

joao-conde (approved):

PR #453, opened 2022-12-14
#429 Add get URL methods

NFSS10:
| - | - |
| --- | --- |
| Dependencies | -- |
| Decisions | - Add `get3dSceneEnvironmentUrl`, `getInitials3dBaseTextureUrl` and `getInitials3dDisplacementTextureUrl` |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -124,6 +124,48 @@ ripe.Ripe.prototype.getMeshUrl = function(options) {
 };
 
 /**
+ * Returns the URL for the environment file of the 3D scene.
+ *
+ * @param {Object} options A map with options, such as:
+ *  - 'brand' - The brand of the model.
+ *  - 'model' - The name of the model.
+ *  - 'version' - The version of the build, defaults to latest.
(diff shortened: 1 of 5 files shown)


src/js/api/brand.js
```diff
@@ -123,6 +123,48 @@ ripe.Ripe.prototype.getMeshUrl = function(options) {
...
+ *
+ * @param {Object} options A map with options, such as:
+ *  - 'brand' - The brand of the model.
+ *  - 'model' - The name of the model.
+ *  - 'version' - The version of the build, defaults to latest.
+ * @returns {String} The URL of the environment for the 3D scene.
+ */
+ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
+    options = this._getMeshOptions(options);
+    return "https://www.dl.dropboxusercontent.com/s/o0v07nn5egjrjl5/studio2.hdr";
```
NFSS10, line 137: The URL is hardcoded because it was discussed at the time to let it have some default value until we have a defined solution to handle the 3D assets

joao-conde: hmm ok

PR #454, opened 2022-12-16
#429 Support metallic, normal and roughness textures

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/429 |
| Dependencies | -- |
| Decisions | - Add `getTextureMapUrl` method<br>- Support metallic, normal and roughness textures<br>- Generalize `CsrRenderedInitials` textures logic |

PR #455, opened 2022-12-20
#429 Use texture_maps endpoint

NFSS10:
| - | - |
| --- | --- |
| Dependencies | - <del>https://github.com/ripe-tech/ripe-sdk/pull/454</del> <br>- <del>https://github.com/ripe-tech/ripe-core/pull/4770</del><br>- <del>https://github.com/ripe-tech/ripe-sdk/pull/473</del> |
| Decisions | Use texture_map endpoint |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -774,10 +774,9 @@ ripe.Ripe.prototype._getMeshOptions = function(options = {}) {
  */
 ripe.Ripe.prototype._getInitials3dOptions = function(options = {}) {
     const brand = options.brand === undefined ? this.brand : options.brand;
-    const model = options.model === undefined ? this.model : options.model;
     const version = options.version === undefined ? this.version : options.version;
     const variant = options.variant === undefined ? this.variant : options.variant;
-    const url = `${this.url}brands/${brand}/models/${model}`;
+    const url = `${this.url}brands/${brand}`;
     const params = {};
     if (version !== undefined && version !== null) {
         params.version = version;
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -769,6 +769,35 @@ ripe.Ripe.prototype._getMeshOptions = function(options = {}) {
     });
 };
 
+const _getProfileValues = (profile, config) => {
+    const $profiles = (config.initials && config.initials.$profiles) || {};
+
+    // obtains the alias corresponding values
+    const aliases = (config.initials && config.initials.$alias) || {};
+    const aliasProfiles = aliases[profile];
+    if (aliasProfiles) {
+        const aliasProfilesArray = Array.isArray(aliasProfiles) ? aliasProfiles : [aliasProfiles];
+        const aliasValues = aliasProfilesArray.reduce((acc, aliasProfile) => {
+            const profileValues = $profiles[aliasProfile] || {};
+            return {
+                ...acc,
+                ...profileValues,
+                "3d": {
+                    ...(acc["3d"] ? acc["3d"] : {}),
+                    ...(profileValues["3d"] ? profileValues["3d"] : {})
+                }
+            };
+        }, {});
+
+        return aliasValues;
+    }
+
+    // obtains the specified profile values
+    const values = $profiles[profile] || {};
+
+    return values;
+};
+
 /**
  * @ignore
  */
@@ -792,9 +821,8 @@ ripe.Ripe.prototype._getInitials3dOptions = function(options = {}) {
     let roughnessTexture = null;
     const config = this.loadedConfig || {};
     const profiles = (config.initials && config.initials.profiles) || [];
-    const $profiles = (config.initials && config.initials.$profiles) || {};
     profiles.forEach(profile => {
-        const values = $profiles[profile] || {};
+        const values = _getProfileValues(profile, config);
         const _3d = values["3d"] || {};
         baseTexture = _3d.base_texture;
         displacementTexture = _3d.displacement_texture;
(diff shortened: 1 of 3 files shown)


src/js/api/brand.js
```diff
@@ -759,6 +769,58 @@ ripe.Ripe.prototype._getMeshOptions = function(options = {}) {
...
+    const $profiles = (config.initials && config.initials.$profiles) || {};
+    profiles.forEach(profile => {
+        const values = $profiles[profile] || {};
+        const _3d = values["3d"] || {};
+        baseTexture = _3d.base_texture;
+        displacementTexture = _3d.displacement_texture;
+        metallicTexture = _3d.metallic_texture;
+        normalTexture = _3d.normal_texture;
+        roughnessTexture = _3d.roughness_texture;
+    });
```
3rdvision, line 804: we should also take into account the aliases in the spec

Back then I implemented something like this

```javascript
        const modelProfileSpec = config.initials.$profiles || {};
        const profileAliases = config.initials.$alias || {};
        const enrichedProfiles = Object.entries(profileAliases).reduce((p, [key, value]) => {
            return profiles.includes(key) ? [...p, ...value] : p;
        }, profiles);
        const modelBaseSpec = modelProfileSpec.base || {};
        const mergedSpec = enrichedProfiles.reduce((p, profile) => {
            return { ...p, ...modelProfileSpec[profile] };
        }, modelBaseSpec);
        console.log("mergedSpec", mergedSpec);
```

This gets the complete spec (mergedSpec) considering the current profile and aliases.

NFSS10: I have a question,
Aren't the `$alias` just that, alias to the profiles names?

What's the idea behind the change, because at it is now what it does is just apply the profiles in their respecting the it's order

3rdvision: The idea is if you have an alias called "report" in the `config.initials.profiles` that is an alias to apply 3 different profiles this will not work as expected - it will try to search for the properties of profile "report" when it should, instead, search for the properties of each profile of the alias "report".

The idea is adding the profiles that are inside a certain alias and **only after** having the (what I called `enrichedProfiles`) should you do what you're doing in your code at `profiles.forEach(...`.

NFSS10: Hmm, what about the profiles order? Doesn't that mean that as it is now, we don't guarantee the profiles order?
I'm asking because `$alias`  is an object

3rdvision: I'm not sure about that but I found a build for it https://github.com/ripe-tech/builds-static/blob/master/config.static/dior_tech/jadore/spec.json - maybe try to play with it and find out what's our current default behavior

NFSS10: Took a deeper dive on how the whole profile thing worked and improved how the values are obtained. Should be good now, take a look

src/js/api/brand.js
```diff
@@ -759,6 +769,94 @@ ripe.Ripe.prototype._getMeshOptions = function(options = {}) {
...
+
+    // obtains the alias corresponding values
+    const aliases = (config.initials && config.initials.$alias) || {};
+    const aliasProfiles = aliases[profile];
+    if (aliasProfiles) {
+        const aliasProfilesArray = Array.isArray(aliasProfiles) ? aliasProfiles : [aliasProfiles];
+        const aliasValues = aliasProfilesArray.reduce((acc, aliasProfile) => {
+            const profileValues = $profiles[aliasProfile] || {};
+            return {
+                ...acc,
```
3rdvision, line 786: rename this acc to something more semantic

3rdvision (commented):
tiny thing

3rdvision (approved):

3rdvision (approved):

src/js/api/brand.js
```diff
@@ -147,7 +147,20 @@ ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
...
-    return null;
+    options = this._getInitials3dOptions(options);
+
+    const textureMap = new Map([
+        ["pattern", options.baseTexture],
+        ["displacement", options.displacementTexture],
+        ["metallic", options.metallicTexture],
+        ["normal", options.normalTexture],
+        ["roughness", options.roughnessTexture]
+    ]);
```
joao-conde, line 158: Any reason why we are using an actual Map instead of our classical JS object?

textureMap = { pattern: options.baseTexture, ....}

Also, seems weird but I have always read objects have faster lookup and are preferred over maps:

https://stackoverflow.com/questions/66931535/javascript-object-vs-map-set-key-lookup-performance
image

joao-conde (changes requested):

src/js/api/brand.js
NFSS10: You are right, using other languages equals doing stuff like this 😅 
Thanks for the heads up!

src/js/api/brand.js
```diff
@@ -147,7 +147,19 @@ ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
...
-    return null;
+    options = this._getInitials3dOptions(options);
+
+    const texture = {
+        pattern: options.baseTexture,
+        displacement: options.displacementTexture,
+        metallic: options.metallicTexture,
+        normal: options.normalTexture,
+        roughness: options.roughnessTexture
+    }[map];
```
joao-conde, line 158: Would it make sense to do

```javascript
const textureMaps = {
   ...
}
const texture = textureMaps[map];
```

joao-conde (changes requested):

src/js/api/brand.js
NFSS10: I did this this way to avoid using switch 🤷‍♂️ 

But if you prefer that I do it this way, is good for me!

joao-conde: Without switch, the same but just having the object in one variable and the result of lookup in another.

joao-conde (approved):

PR #463, opened 2023-01-05
#429 Support loading of external fonts

NFSS10:
| - | - |
| --- | --- |
| Dependencies | <del>https://github.com/ripe-tech/ripe-sdk/pull/455</del> |
| Decisions | Add `getFontUrl()` and `_getFontOptions()` methods<br> Load font based on config |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -752,6 +752,9 @@ ripe.Ripe.prototype._getFontOptions = function(options = {}) {
     if (variant !== undefined && variant !== null) {
         params.variant = variant;
     }
+    if (options.weight !== undefined && options.weight !== null) {
+        params.weight = options.weight;
+    }
     return Object.assign(options, {
         url: url,
         method: "GET",
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -124,6 +124,22 @@ ripe.Ripe.prototype.getMeshUrl = function(options) {
 };
 
 /**
+ * Returns the URL where the font can be retrieved.
+ *
+ * @param {Object} options A map with options, such as:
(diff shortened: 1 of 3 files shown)


src/js/api/brand.js
```diff
@@ -123,6 +123,22 @@ ripe.Ripe.prototype.getMeshUrl = function(options) {
...
+ *
+ * @param {Object} options A map with options, such as:
+ *  - 'brand' - The brand of the model.
+ *  - 'model' - The name of the model.
+ *  - 'version' - The version of the build, defaults to latest.
+ * @returns {String} The URL of the specified initials texture.
+ */
+ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
+    options = this._getFontOptions(options);
+    format = format || "ttf";
```
3rdvision, line 137: I'd simplify like this... what do you think?

```suggestion
ripe.Ripe.prototype.getFontUrl = function(name, format = "ttf", options) {
    options = this._getFontOptions(options);
```

3rdvision (commented):
Small thing

src/js/api/brand.js
NFSS10: I usually prefer to have as it is, becase you can do 
`getFontUrl ("Example", undefined, { stuff here })`
and it will always preserve the default, otherwise you would have to spcify the default

NFSS10: But I can change if you think I should

3rdvision: it was only a matter of simplification but indeed there's an advantage for that usecase

3rdvision (approved):

joao-conde (approved):

3rdvision (approved):

PR #468, opened 2023-01-19
#4778 Fix tests

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-core/issues/4778 |
| Decisions | - Fix failing tests by using `imagesDefault` set to false <br> - Add support for param `imagesDefault` (RIPE Core `images_default`)|

### DIFF
--- a/src/js/api/order.js
+++ b/src/js/api/order.js
@@ -2456,6 +2456,7 @@ ripe.Ripe.prototype._importOrder = function(ffOrderId, options = {}) {
     const notes = options.notes === undefined ? null : options.notes;
     const images = options.images === undefined ? null : options.images;
     const safe = options.safe === undefined ? null : options.safe ? "1" : "0";
+    const imagesDefault = options.imagesDefault === undefined ? null : options.imagesDefault ? 1 : 0;
 
     const url = `${this.url}orders/import`;
     const contents = {};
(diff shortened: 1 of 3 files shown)


src/js/api/order.js
```diff
@@ -2456,6 +2456,7 @@ ripe.Ripe.prototype._importOrder = function(ffOrderId, options = {}) {
     const notes = options.notes === undefined ? null : options.notes;
     const images = options.images === undefined ? null : options.images;
     const safe = options.safe === undefined ? null : options.safe ? "1" : "0";
+    const imagesDefault = options.imagesDefault === undefined ? null : options.imagesDefault ? 1 : 0;
```
joao-conde, line 2459: Just like "safe" above, dont we want this to be "1" and "0" (strings instead of integers)

joao-conde (changes requested):

src/js/api/order.js
3rdvision: RIPE Core is not doing proper casting... "0" casts to `True`

3rdvision: Also... We can't elegantly fix it in the core side since all it's doing is `images_default = contents.get("images_default", True)` to later verify with `if not images and images_default:`.
This is the reason why an `images_default` of value `0` is "misinterpreted" as `True`

joao-conde: How is it working for "safe" currently?

3rdvision: "safe" is not inside `contents` so we can do the proper casting with `safe = self.field("safe", True, cast=bool)`

joao-conde (approved):

PR #469, opened 2023-01-25
#429 Support post render callback

NFSS10:
| - | - |
| --- | --- |
| Dependencies | -- |
| Decisions | Allows setting a callback to be ran immediately after the render. This is useful for cases where, for example, we want immediate access to the renderer frame buffer |

### DIFF
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -1209,6 +1209,7 @@ ripe.ConfiguratorCsr.prototype._render = function() {
     if (!this.camera) throw new Error("Camera not initiated");
     if (!this.renderer) throw new Error("Renderer not initiated");
     this.renderer.render(this.scene, this.camera);
+    this._onPostRender();
 };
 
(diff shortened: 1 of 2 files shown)


src/js/visual/configurator-csr.js
```diff
@@ -528,6 +529,22 @@ ripe.ConfiguratorCsr.prototype.prcFrame = async function() {
...
+ * Sets the callback function for the post render call.
+ *
+ * @param {Function} callback The function to be called.
+ */
+ripe.ConfiguratorCsr.prototype.setPostRender = function(callback) {
+    this._postRenderCallback = callback;
+};
+
+/**
+ * Clears the the post render callback registry.
```
3rdvision, line 542:
```suggestion
 * Clears the post render callback registry.

```

:eye:

3rdvision (changes requested):
small thing

3rdvision (approved):

PR #484, opened 2023-02-17
#1098 Add params to `ready` event in configurator csr and prc

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white/issues/1098 |
| Decisions | - Add params to `ready` event in configurator csr  and prc, allowing for any loaders to catch this event. This is because `ready` event is triggered in a lot of places and not specific enough to use as a loader bind. |

### DIFF
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -58,6 +58,7 @@ ripe.ConfiguratorCsr.prototype.init = function() {
     ripe.Visual.prototype.init.call(this);
 
     // options variables
+    this.id = this.options.id || null;
     this.width = this.options.width || null;
     this.height = this.options.height || null;
     this.size = this.options.size || null;
@@ -1740,7 +1741,7 @@ ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
         await this.flushPending(true);
 
         self.loading = false;
-        this.trigger("ready", { id: "configurator_csr" });
+        this.trigger("ready", { id: this.id, origin: "configurator-csr" });
     };
 
     // runs synchronously or asynchronously depending on how the CSR configurator was setup
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -58,7 +58,7 @@ ripe.ConfiguratorCsr.prototype.init = function() {
     ripe.Visual.prototype.init.call(this);
 
     // options variables
-    this.id = this.options.id ?? 0;
+    this.id = this.options.id || 0;
     this.width = this.options.width || null;
     this.height = this.options.height || null;
     this.size = this.options.size || null;
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -58,7 +58,7 @@ ripe.ConfiguratorCsr.prototype.init = function() {
     ripe.Visual.prototype.init.call(this);
 
     // options variables
-    this.id = this.options.id || null;
+    this.id = this.options.id ?? 0;
     this.width = this.options.width || null;
     this.height = this.options.height || null;
     this.size = this.options.size || null;
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -1740,8 +1740,7 @@ ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
         await this.flushPending(true);
 
         self.loading = false;
-        this.trigger("ready");
-        this.trigger("post_config", config);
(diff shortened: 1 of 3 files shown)


src/js/visual/configurator-csr.js
```diff
@@ -1741,6 +1741,7 @@ ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
 
         self.loading = false;
         this.trigger("ready");
+        this.trigger("post_config", config);
```
NFSS10, line 1744: Won't this do a infinite loop?
This method is run as a callback for `post_config` event and not doing the config itself.

NFSS10 (changes requested):

src/js/visual/configurator-csr.js
3rdvision: I get what you're saying because of 
```js
ripe.ConfiguratorCsr.prototype._registerConfigHandlers = function() {
    this.owner.bind("pre_config", (brand, model, options) => this._onPreConfig(this));
    this.owner.bind("post_config", config => this._onPostConfig(this, config)); // <-- this
};
```

It doesn't do an infinite loop because `this.trigger` triggers a configurator event which is different from `this.owner.trigger`.

I do recognize the name of the event is kind of generic but I wasn't sure of the best option since I also wanted to keep it generic and not something like `post_config_csr`.

Do you have any suggestion?

NFSS10: Hmm 🤔 
So the idea is to know when the configurator is done loading right? If that's the case I would use the `ready` event and just make it have params and one of the params being a ID of the stuff we are tracking

3rdvision: Added `{ id: "configurator_csr" }` inspired by the name of the context.
Now I'm not sure this is the best structure. Maybe I'd prefer `{ origin: "configurator_csr" }` instead. What do you say?

src/js/visual/configurator-csr.js
```diff
@@ -1740,8 +1740,7 @@ ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
         await this.flushPending(true);
 
         self.loading = false;
-        this.trigger("ready");
-        this.trigger("post_config", config);
+        this.trigger("ready", { id: "configurator_csr" });
```
NFSS10, line 1743: I suggest allowing to specify the ID as an optional parameter here: https://github.com/ripe-tech/ripe-sdk/blob/3bf809a5a6fee6d905fbfc007a53d81cd8f639cc/src/js/visual/configurator-csr.js#L60

This way it will make it possible to differantiate the configurators. Maybe also add to PRC configurator?

NFSS10 (changes requested):

src/js/visual/configurator-csr.js
NFSS10: I guess it depends on what is the purpose of the variable. If it's to inform just origin of the event, maybe it makes sense. Even a specified event for that. If it's to differentiate configurators, I guess an ID is better

NFSS10 (changes requested):

src/js/visual/configurator-csr.js
```diff
@@ -58,6 +58,7 @@ ripe.ConfiguratorCsr.prototype.init = function() {
     ripe.Visual.prototype.init.call(this);
 
     // options variables
+    this.id = this.options.id ?? 0;
```
NFSS10: I think we should maintain the behavior consistent, but cool trick 😄

I would also suggest to have maintain this javascripty and let the ID be a string, gives more flexability

NFSS10 (changes requested):

NFSS10:
@3rdvision Also update PR info like title and stuff

3rdvision:
> @3rdvision Also update PR info like title and stuff

@NFSS10 done

NFSS10:
@3rdvision Just a thing that I noticed.

Isn't there a conflict on initializing the id with this? https://github.com/ripe-tech/ripe-sdk/blob/3bf809a5a6fee6d905fbfc007a53d81cd8f639cc/src/js/visual/visual.js#L50

3rdvision:
> @3rdvision Just a thing that I noticed.
> 
> Isn't there a conflict on initializing the id with this?
> 
> https://github.com/ripe-tech/ripe-sdk/blob/3bf809a5a6fee6d905fbfc007a53d81cd8f639cc/src/js/visual/visual.js#L50

Good call. I think they are different because `this.id` is the `ripeInstance` id and inside the configurator scope `this.id` is the `configurator` id.

I double checked and it's the case:
image

Issue #479, opened 2023-02-14
CSR performance optimizations


NFSS10:
## Rationale

There are some performance gains to be made in the CSR strategy, some of which quick wins others that can come from the addition of new features/settings or from a more calmly review of the code.

## Description

Review and improve the CSR performance. This may be by, when possible and if applicable, using parallelization, reusing resources, avoiding unnecessary remote calls, avoiding unnecessary usage/loading of resources, etc.

- [ ] Use compressed textures
- [ ] Run initials renderer in parallel if possible
- [ ] Check the possibility of using workers
... maybe more???

PR #485, opened 2023-02-21
#479 Support computation of initials config by adding `initialsConfig` method

NFSS10:
| - | - |
| --- | --- |
| Issue |https://github.com/ripe-tech/ripe-sdk/issues/479 |
| Dependencies | -- |
| Decisions | - Add initialsConfig() method which runs the appy profiles logic locally. It's a implementation that matches the one done here: https://github.com/ripe-tech/ripe-compose/blob/40436c5207b01fdff100e6d2425018feb453c731/src/ripe_compose/util/configurator.py#L2213. The reason it's done is to support a way to compute the initials with the profiles applicated without having to do a remote call<br>- Add `initialsConfig()` tests |

### DIFF
--- a/src/js/base/config.js
+++ b/src/js/base/config.js
@@ -57,7 +57,7 @@ ripe.Ripe.prototype.initialsConfig = function(config, profiles = []) {
     }
 
     const profilesValues = initials.$profiles || {};
-    const alias =  initials.$alias || {};
+    const alias = initials.$alias || {};
 
     const profilesFinal = [];
     baseProfiles.reverse().forEach(profile => {
@@ -67,13 +67,13 @@ ripe.Ripe.prototype.initialsConfig = function(config, profiles = []) {
             const values = profilesValues[p];
             if (!values) return;
 
-            initials = { ...initials, ...values }
+            initials = { ...initials, ...values };
             profilesFinal.push(p);
         });
     });
 
     const initialsRoot = initials.$root || {};
-    initials = {...initials, ...initialsRoot };
+    initials = { ...initials, ...initialsRoot };
 
     initials.profiles = profilesFinal;
 
--- a/src/js/base/config.js
+++ b/src/js/base/config.js
@@ -52,6 +52,9 @@ ripe.Ripe.prototype.hasStrategy = function(strategy) {
  * `profiles` argument is provided those profiles will be also used to compute the
  * initials config.
  *
+ * This method is implements the same logic as `initials_config` in the composition
+ * engine this should be kept in sync with that implementation.
+ *
  * @param {Object} config The model's config.
  * @param {Array} profiles The list of profiles to use.
  * @returns {Object} The computed initials config based on the config profiles and
--- a/src/js/base/config.js
+++ b/src/js/base/config.js
@@ -77,15 +77,29 @@ ripe.Ripe.prototype.initialsConfig = function(config, profiles = []) {
             const values = profilesValues[p];
             if (!values) return;
 
-            initials = { ...initials, ...values };
+            initials = this._initialsUpdate(initials, values);
             profilesFinal.push(p);
         });
     });
 
     const initialsRoot = initials.$root || {};
-    initials = { ...initials, ...initialsRoot };
+    initials = this._initialsUpdate(initials, initialsRoot);
 
     initials.profiles = profilesFinal;
 
     return initials;
+};
+
+/**
+ * Retrieves the updated initials config by merging the initials config values
+ * with other initials config values.
+ *
+ * @param {Object} initials The base initials config values.
+ * @param {Object} values The initials config values to be applied.
+ * @returns {Object} The update initials config.
+ * 
+ * @private
+ */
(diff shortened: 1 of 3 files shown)


src/js/base/config.js
```diff
@@ -46,3 +46,60 @@ ripe.Ripe.prototype.hasStrategy = function(strategy) {
...
+ * Retrieves the initials config with the specified profiles in the config. If the
+ * `profiles` argument is provided those profiles will be also used to compute the
+ * initials config.
+ *
+ * @param {Object} config The model's config.
+ * @param {Array} profiles The list of profiles to use.
+ * @returns {Object} The computed initials config based on the config profiles and
+ * specified profiles.
+ */
+ripe.Ripe.prototype.initialsConfig = function(config, profiles = []) {
```
3rdvision, line 60: Since we are not removing the remote method I'm assuming we maintaining the `getInitialsConfigP` method to be used as an alternative remote. The so called `remote = true` approach we discussed.

test/js/base/config.js
```diff
@@ -64,4 +64,97 @@ describe("Config", function() {
             assert.strictEqual(instance.hasInitialsRadius(), true);
         });
     });
+
+    describe("#initialsConfig()", function() {
+        it("should be able to retrieve the initials config with the applied profiles", () => {
```
3rdvision, line 69: Would be nice if we had a test to compare the result of local RIPE SDK `initialsConfig()` against remote RIPE Core `getInitialsConfigP()`. I think considering the risk and all the discussion it's important to add some confidence that the local method is in sync with the remote method.

test/js/base/config.js
```diff
@@ -64,4 +64,97 @@ describe("Config", function() {
             assert.strictEqual(instance.hasInitialsRadius(), true);
         });
     });
+
+    describe("#initialsConfig()", function() {
```
3rdvision, line 68: This was ported from by https://github.com/ripe-tech/ripe-compose/blob/08b211a1c7968cdb655ce1c4ffbe14ecbb15dd80/src/ripe_compose/test/util/configurator.py#L548 correct?

src/js/base/config.js
```diff
@@ -46,3 +46,60 @@ ripe.Ripe.prototype.hasStrategy = function(strategy) {
     const strategies = (this.loadedConfig && this.loadedConfig.strategies) || ["prc"];
     return strategies.includes(strategy);
 };
+
+/**
+ * Retrieves the initials config with the specified profiles in the config. If the
+ * `profiles` argument is provided those profiles will be also used to compute the
+ * initials config.
+ *
```
3rdvision, line 54: Please add a comment mentioning that this logic should be in sync with composition engine's logic when computing the initials config - currently `initials_config`

3rdvision (commented):
Some improvements as previously discussed.

test/js/base/config.js
NFSS10: Yes

NFSS10:
Linking discussion thread: https://platforme.slack.com/archives/C04PEFAK4BW/p1676996739770149

src/js/base/config.js
NFSS10: Yes, this is just an alternative that runs locally instead of doing the remote call

src/js/base/config.js
```diff
@@ -46,3 +46,65 @@ ripe.Ripe.prototype.hasStrategy = function(strategy) {
...
     return strategies.includes(strategy);
 };
+
+/**
+ * Retrieves the initials config with the specified profiles in the config. If the
+ * `profiles` argument is provided those profiles will be also used to compute the
+ * initials config.
+ *
+ * This method is implements the same logic as `initials_config` in the composition
+ * engine this should be kept in sync with that implementation.
```
3rdvision, line 56:
```suggestion
 * This method implements the same logic as `initials_config` in the composition
 * engine and should be kept in sync with that implementation.
```

test/js/base/config.js
```diff
@@ -64,4 +64,138 @@ describe("Config", function() {
...
+            ]);
+            assert.strictEqual(initialsConfig.font_family, "Silver");
+        });
+
+        it("should be able to match the config of the remote initials config call", async () => {
+            const instance = await new ripe.Ripe("swear", "vyner", { noBundles: true });
+            await instance.isReady();
+
+            let remote = await instance.getInitialsConfigP();
+            let local = instance.initialsConfig(instance.loadedConfig);
```
3rdvision, line 165: Can we call these methods with the exact same input?
I think this is nice to enforce that these, given the same input, should return the same output.

I'm suggesting something like:

```suggestion
            const ripeConfig = instance.loadedConfig;

            let remote = await instance.getInitialsConfigP(ripeConfig);
            let local = instance.initialsConfig(ripeConfig);
```

NFSS10: No, we the `getInitialsConfigP` doesn't work like that

3rdvision (approved):

Issue #478, opened 2023-02-14
Review and improve documentation


NFSS10:
## Rationale

The RIPE-SDK documentation is lacking in some aspects such as examples and demos that could help an external person to easily dive in and start using it.

## Description

Analyze and check what can be improved with the documentation and maybe create some demos and improve the examples on how to use the SDK, namely the configurator and each of the strategies (PRC and CSR)

NFSS10:
I have a personal repository with some simple examples of RIPE-SDK: https://github.com/NFSS10/ripe-sdk-examples
It may be useful!

PR #492, opened 2023-03-01
#478 Add CSR alt documentation

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/478 |
| Decisions | - Add CSR base documentation |
| Animated GIF | image |


documentation/vercel.json
```diff
@@ -0,0 +1,6 @@
+{
+    "installCommand": "npm install",
+    "buildCommand": "npm run vercel-build",
+    "outputDirectory": "src/.vuepress/dist",
+    "framework": null
+}
```
NFSS10, line 6: Should I be able to see a deployment of this or is this added by mistake?

NFSS10: Also I may be wrong, but I think we can have vercel deployement without the `vercel.json`

NFSS10 (changes requested):
@3rdvision I think renaming the folder from `documentation` to `docs` would better, WDYT?

3rdvision:
> @3rdvision I think renaming the folder from `documentation` to `docs` would better, WDYT?

True. The problem was naming collision as ./docs is already used by the Dockerfile

documentation/vercel.json
3rdvision: This allows to defined the vercel config here instead of having to config in their interface.
It's not automatic because it's not the default in this case

NFSS10: Ok, but I think we should maintain this agnostic to any service this may use. If it's not needed, it should not be added to not polute the repo.

3rdvision: We're already doing it for e.g in [ripe-util-vue](https://github.com/ripe-tech/ripe-util-vue/blob/master/vercel.json)
I think it should stay here, this way no one has to remember that we need to pick the "Other" framework, specify the `src/.vuepress/dist` as the ouput directory and so on... 

Also, if we change the path of `dist` folder we can just update it here in the `outputDirectory` instead of having to change it in vercel.

In short this allows to ease vercel setup and maintainability

NFSS10: That's why I'm talking about this. That file on util-vue is a mistake, that isn't even supported right now, that setup is deprecated, that's why I'm talking about this to avoid doing stuff like that. If it isn't needed, it shouldn't be added.

At most we should add documentation on how to setup in the services we are currently using and not make it part of the project. It will make it more maintainable as we are not just having a random file that doesn't work (or no one knows what it's actually doing). The setup will be done one time, so I think it's a good trade off.
WDYT?

3rdvision: > That file on util-vue is a mistake, that isn't even supported right now
That file is still supported, [it's mentioned in vercel docs](https://vercel.com/docs/concepts/projects/project-configuration) and it's needed - if you remove it ripe-util-vue vercel deploy will fail.

>  The setup will be done one time
The same argument applies to `vercel.json` being added in ripe-util-vue and [many other](https://github.com/search?q=org%3Aripe-tech%20vercel.json&type=code).
The setup can be done in one time but might break in the future and you have no visibility on it. In here, you can control the configuration of the project by code and version it.

I don't think this is poluting but following good practice. We can find `vercel.json` files in all kinds of repos, including our own.

This vercel.json file will help developers understand what configs **that are not the default*** need to be done in order for the deployment to work in vercel.

NFSS10: Yes the file exists for custom setup I know that. I was talking about the things we use are already deprecated: (example: https://vercel.com/docs/concepts/projects/project-configuration#builds)

The thing about breaking in the future is why I recommended documentation about the setup of the service that is deploying this. As in, instead of having a `vercel.json` that doesn't have nothing special and that have things that can be done via the interface, we have an "how to" which gives more information on the setup (the what and the whys).

My train of thought is more in line with stuff like we changing to Netlify or some other random service, or just Vercel deprecating stuff again. The ideia is to detach from things that aren't necessarly crucial to the repo and avoid polluting it as much as possible (one example of similar stuff is `.vscode`, `.idea`, etc)

I also think it will be helpfull if in the future someone needs to update the setup, that way they can see why stuff was done in X or Y way instead of having a "plug and play black box". One example that made me thought of this was our setup in ripe-api-docs.

But maybe I'm thinking to much about this. I just wanted to have the thought out there so I'm ok with having this file if in the end is what you prefer 🙂

3rdvision: >  I was talking about the things we use are already deprecated

True, indeed we shouldn't be using some of those properties. In this case I'm not including any deprecated or non-recommended properties in this `vercel.json`.

> My train of thought is more in line with stuff like we changing to Netlify or some other random service

That is a valid point but I think the `vercel.json` might actually help us understand how the setup is done even in Vercel's specification.
Considering an unlikely scenario in which Vercel vanishes from the internet or we lose access to it and want to port the deployment to somewhere else, the configuration would be still accessible to us in this `vercel.json`. Even though it's Vercel's spec, it's simple enough to understand how the setup is being done and port to Netlify or similar if necessary. In the same scenario where there was no `vercel.json` that setup information would only be stored in Vercel's servers which I think is worse in any case.

Indeed there could be a more agnostic way of deploying config specification but as of now I think there isn't a standard to this. I also think that this is such a simple deployment that might not even be the best example.

We should be making this decision and apply it to more repositories because we don't have a standard for it. Maybe a `Dockerfile`, maybe something else.

In this case I think it's worth maintaining the `vercel.json` and later on if we decide to create an agnostic setup config, apply it to all repositories.

NFSS10: It's fine by me then! 🙂

NFSS10 (approved):

PR #494, opened 2023-03-01
#493 Improve CSR deinitialization

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/493
| Decisions | - Improve de-initialization logic by adding missing `_destroyConfig`, assigning proper event callback to allow proper unbinding match and avoid event handler duplication by making sure previous handlers are unbinded first before binding.  |

### DIFF
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -160,6 +160,8 @@ ripe.ConfiguratorCsr.prototype.init = function() {
 ripe.ConfiguratorCsr.prototype.deinit = async function() {
     this._deinitCsr();
 
+    this._unregisterConfigGlobalHandlers();
+
     while (this.element && this.element.firstChild) {
         this.element.removeChild(this.element.firstChild);
     }
@@ -1167,8 +1169,6 @@ ripe.ConfiguratorCsr.prototype._deinitCsr = function() {
     this._destroyDebug();
     this._destroyInitialsResources();
     this._destroyScene();
-
-    this._unregisterConfigGlobalHandlers();
 
     if (this.renderer) {
         this.renderer.dispose();


3rdvision:
> That behaviour is intended

What behavior is intended?

There are several things that need to be fixed:

* adding missing _destroyConfig that calls `_unregisterConfigHandlers`  - currently `_unregisterConfigHandlers` is not being used 
* assigning proper event callback to allow proper unbinding match (the current implementation doesn't unbind anything)
* avoid event handler duplication by making sure previous handlers are unbinded first before binding - this is more specific for chaotic `post_config` which is ripe-white but an extra guard adds more confidence that we always have only 1 callback for each bind

NFSS10:
>  adding missing `_destroyConfig` that calls `_unregisterConfigHandlers`  - currently `_unregisterConfigHandlers` is not being used

It's not missing anything, the `_unregisterConfigHandlers()` isn't called because there is no behaviour that handles the `deinit()` midway the configurator usage. This line handles that deinit ((line 164)): https://github.com/ripe-tech/ripe-sdk/blob/ed0607fc0613f3de98f3b73ebb2a9d85537e49df/src/js/visual/configurator-csr.js#L164
<br>

> assigning proper event callback to allow proper unbinding match (the current implementation doesn't unbind anything)

It does, the binds are in the owner. That logic is in the observable and the configurator inherits that logic (`Obersvable` -> `Interactable` -> `Visual` -> `Configurator`). As I said above, deinit handles that.
<br>

> avoid event handler duplication by making sure previous handlers are unbinded first before binding - this is more specific for 
chaotic `post_config` which is ripe-white's case but an extra guard adds more confidence that we always have only 1 callback for each bind

Don't know what this means but it shouldn't as the config handlers are only registered when you init a configurator thus that only happens once per configurator and for the initials bindings it's already handling as it reconstructs and reregisters itself in mid usage.
<br>

> adding missing `_destroyConfig` that calls `_unregisterConfigHandlers`  - currently `_unregisterConfigHandlers` is not being used

Same as the first point.

3rdvision:
I'll separate by points otherwise this will be hard to track.

---

1- 

> It's not missing anything, the _unregisterConfigHandlers() isn't called because there is no behaviour that handles the deinit() midway the configurator usage. This line handles that deinit ((line 164)): 

`_unregisterConfigHandlers` is dead code. It's not being called anywhere. `ripe.Visual.prototype.deinit.call(this);` does not call it. I'm not sure what you're suggesting but `deinit` should unregister all the event bindings and it's not doing it.  Feel free to suggest another fix.

---

2-

> It does, the binds are in the owner. That logic is in the observable and the configurator inherits that logic (`Obersvable` -> `Interactable` -> `Visual` -> `Configurator`). As I said above, deinit handles that.

When doing `unbind` with `this.owner && this.owner.unbind("initials_extra", this._onInitialsExtraEvent);`, we call this `removeCallback`  which then does [this check](https://github.com/ripe-tech/ripe-sdk/blob/ed0607fc0613f3de98f3b73ebb2a9d85537e49df/src/js/base/observable.js#L65). The check never finds that method because it wasn't that method that was binded, it was a new arrow function of that method but that has a difference reference.

The way it's currently implemented in `configurator-csr`, it doesn't find any reference of `this._onInitialsExtraEvent` because the actual function that was saved in the callback array was a new reference for an arrow function done with

```
  this.owner.bind("initials_extra", (initialsExtra, params) =>
      this._onInitialsExtraEvent(this, initialsExtra, params)
  );
```

The reference that is saved when receiving `(initialsExtra, params) =>       this._onInitialsExtraEvent(this, initialsExtra, params)` in the args is different than receiving `this._onInitialsExtraEvent`.

Also, when I reviewed this at first, I didn't caught but we should follow the same methodology done in other places like [here](https://github.com/ripe-tech/ripe-sdk/blob/ed0607fc0613f3de98f3b73ebb2a9d85537e49df/src/js/plugins/sync.js#L74). The bind is done the correct way - it binds the callback, saves it globally inside `this._partBind` and later does `this.owner.unbind("part", this._partBind);`.

to better understand this you can try in your browser to do
```
func1 = () => {}
func1 === (() => func1) // this returns false because the reference of a new arrow func is not the same as the function itself
```

---

> Don't know what this means but it shouldn't as the config handlers are only registered when you init a configurator thus that only happens once per configurator and for the initials bindings it's already handling as it reconstructs and reregisters itself in mid usage.

It should only happen once per configurator, unfortunately that is not the case. This was a very tricky one to debug and understand. The weird behavior that happens is because `post_config` is called twice in some scenarios and because of it 2 handlers will be binded and only one will be unbinded after deinit. How do you suggest do fix this then?

Part of the issue is because CSR only initializes after a `post_config` event. And in some scenarios you may already have a model configurated before binding the configurator. This is the case for ripe-white.

NFSS10:
`_unregisterConfigHandlers()` being dead code is a yes in the sense of it not being used, but a no because it's not being used because at the current state there is no behaviour that needes to call it mid lifecycle of the CSR configurator so it's just an unused method waiting to be called when needed.

What I'm saying is that the Visual.deinit will reach this: https://github.com/ripe-tech/ripe-sdk/blob/7a07234131d81c2445b1aae9b2f8a66a755d80ae/src/js/base/observable.js#L33
and clear all callbacks as it's supposed to.

About what you are saying, to see if I'm understanding correctly. You are essently saying that the Visual.deinit is not correctly unbinding the owner binds?
We need to know exactly the problem because we had some issues a few years ago related to this area (one of those related to thumbnails not being properly discarded) and the issue was because essentially the SDK logic is not well built and some of the components where using it's logic incorrectly.

It's important to know the situation because It's not the Configurator responsibility to explicitly clear the `owner` binds on `deinit()` that's why we call the parent method. Doing that would be redundant or a symptom of a bug somewhere so we have to make sure.
The only cases where we want to explicitly do `owner` unbinds is if we have some kind of behaviour that will need to do it mid cycle of the configurator lifecycle and it seems is not the case here.

NFSS10:
Ok, scratch the last part, it's being used in `_deinitCsr` and that's used mid lifecycle, my bad!

src/js/visual/configurator-csr.js
```diff
@@ -1771,10 +1779,12 @@ ripe.ConfiguratorCsr.prototype._registerHandlers = function() {
  * @ignore
  */
 ripe.ConfiguratorCsr.prototype._registerInitialsHandlers = function() {
-    this.owner.bind("initials", (initials, engraving, params) =>
+    // unregister previous config handlers to avoid duplicate handlers
+    this._unregisterInitialsHandlers();
```
NFSS10, line 1783: Don't call the unregister method here. If you need to do it here, it's a symptom of a bug

src/js/visual/configurator-csr.js
```diff
@@ -1783,22 +1793,24 @@ ripe.ConfiguratorCsr.prototype._registerInitialsHandlers = function() {
...
 };
 
 /**
  * @ignore
  */
 ripe.ConfiguratorCsr.prototype._registerConfigHandlers = function() {
-    this.owner.bind("pre_config", (brand, model, options) => this._onPreConfig(this));
-    this.owner.bind("post_config", config => this._onPostConfig(this, config));
+    // unregister previous config handlers to avoid duplicate handlers
+    this._unregisterConfigHandlers();
```
NFSS10, line 1805: Don't call the unregister method here. If you need to do it here, it's a symptom of a bug

src/js/visual/configurator-csr.js
```diff
@@ -1783,22 +1793,24 @@ ripe.ConfiguratorCsr.prototype._registerInitialsHandlers = function() {
...
+    this._onPreConfigEventCallback = this.owner.bind("pre_config", (brand, model, options) => this._onPreConfig(this));
+    this._onPostConfigEventCallback = this.owner.bind("post_config", config => this._onPostConfig(this, config));
```
NFSS10, line 1807: Make the name shorter, like `_onPreConfigBind`, and add the variable declaration to here: https://github.com/ripe-tech/ripe-sdk/blob/b327d752b9578f542abace9612f4276d7dc6fc27/src/js/visual/configurator-csr.js#L95

src/js/visual/configurator-csr.js
```diff
@@ -1771,10 +1779,12 @@ ripe.ConfiguratorCsr.prototype._registerHandlers = function() {
...
  */
 ripe.ConfiguratorCsr.prototype._registerInitialsHandlers = function() {
-    this.owner.bind("initials", (initials, engraving, params) =>
+    // unregister previous config handlers to avoid duplicate handlers
+    this._unregisterInitialsHandlers();
+    this._onInitialsEventCallback = this.owner.bind("initials", (initials, engraving, params) =>
         this._onInitialsEvent(this, initials, engraving, params)
     );
-    this.owner.bind("initials_extra", (initialsExtra, params) =>
+    this._onInitialsExtraEventCallback = this.owner.bind("initials_extra", (initialsExtra, params) =>
```
NFSS10, line 1787: Make the name shorter, like `_onInitialsExtraBind`, and add the variable declaration to here: https://github.com/ripe-tech/ripe-sdk/blob/b327d752b9578f542abace9612f4276d7dc6fc27/src/js/visual/configurator-csr.js#L95

I would also take the opportunity to to shorten the name from `_onInitialsExtraEvent´ to `_onInitialsExtra` to match the `_onPostConfig` (or the other way around)

src/js/visual/configurator-csr.js
```diff
@@ -1160,6 +1167,7 @@ ripe.ConfiguratorCsr.prototype._initCsr = function() {
 ripe.ConfiguratorCsr.prototype._deinitCsr = function() {
     this._destroyDebug();
     this._destroyInitialsResources();
+    this._destroyConfig();
```
NFSS10, line 1170: Hmm, this shouldn't be here, it's not CSR related, as in, it's not related to the scene building and Three.js resources stuff.

Also this is what I'm talking about, this shouldn't affect anything as the `_deinitCsr` is only called on the `deinit()` and that will clear all `owner` binds

src/js/visual/configurator-csr.js
```diff
@@ -949,6 +949,13 @@ ripe.ConfiguratorCsr.prototype._destroyInitialsResources = function() {
     this._deinitCsrRenderedInitials();
 };
 
+/**
+ * Completely cleanup and destroy CSR initials.
+ */
```
NFSS10, line 954: It cleanups config stuff, not CSR initials

NFSS10 (changes requested):
https://github.com/ripe-tech/ripe-sdk/pull/494#issuecomment-1452066603

Scratch that again, as I was right too 😅 , `_deinitCsr()` is only called when `deInit()` is called on the configurator lifecycle, thus the points still stand.

src/js/visual/configurator-csr.js
3rdvision: The bug cause is `_onPostConfig ` being called several times. This results in binding several initials handlers. Later those initials handlers are not removed because we're only saving the callback of 1 of them.

src/js/visual/configurator-csr.js
3rdvision: let's resolve this [here](https://github.com/ripe-tech/ripe-sdk/pull/494#discussion_r1123505246)

3rdvision:
>  _deinitCsr() is only called when deInit() is called on the configurator lifecycle, thus the points still stand.

I'll double check this in RIPE White context because this was not being the case

src/js/visual/configurator-csr.js
3rdvision: > Also this is what I'm talking about, this shouldn't affect anything as the _deinitCsr is only called on the deinit() and that will clear all owner binds

Can you explain how the `deinit()` clears all owner binds?

I don't see that happening and I couldn't find the logic of where this happens.

Here is a video of flawless model swap in RIPE White using CSR:

https://user-images.githubusercontent.com/24736423/222774339-f6ff6a91-f451-4d75-9357-5a71c3dc7b8b.mp4

Here is another video with the same setup the only difference being that I removed the line of code that calls `this._destroyConfig();`:

https://user-images.githubusercontent.com/24736423/222774883-1b219828-4158-48b0-a1b6-e22e1935c356.mp4

This is to show you that these binds:
```

    this._onPreConfigBind = this.owner.bind("pre_config", (brand, model, options) => this._onPreConfig(this));
    this._onPostConfigBind = this.owner.bind("post_config", config => this._onPostConfig(this, config));
```

need to be properly unbinded, specially after that configurator stops to exist for some reason and at this current implementation they aren't.

What's your suggestion on fixing this considering that now we know that we need to do these unbindings when doing `_deinitCsr`?

NFSS10: Here is a simple example:

The button just calls configurator `.deinit()` and you can see that the callbacks are all cleared because `callbacks` is set to `null` in the `Observable` `deinit()` call


https://user-images.githubusercontent.com/22588915/223089000-651ea97c-ca82-46a9-9892-fa2897b99e4a.mp4

Change file format to `.html`
[csr.txt](https://github.com/ripe-tech/ripe-sdk/files/10897257/csr.txt)

src/js/visual/configurator-csr.js
NFSS10: But it being called multiple times is maybe a symptom of some bug of the caller as this is just used to register. If you need to unregister before registering again, then the caller of `_registerInitialsHandlers()` should call `_unregisterInitialsHandlers()`, It's not `_registerInitialsHandlers()` responsibility to call `_unregisterInitialsHandlers()`

3rdvision:
@NFSS10 I pushed the changes we discussed in person.

Let me know if there's anything missing.

src/js/visual/configurator-csr.js
```diff
@@ -1703,6 +1717,7 @@ ripe.ConfiguratorCsr.prototype._onPreConfig = function(self) {
  * @ignore
  */
 ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
+    if (!this.owner.hasStrategy("csr")) return;
```
NFSS10, line 1720: Shouldn't this not be here? This just tells me there is a bug somewhere, otherwise this kind of guard should be anywhere and everywhere which is why this is smelling like a bug somewhere.

src/js/visual/configurator-csr.js
```diff
@@ -1160,6 +1173,7 @@ ripe.ConfiguratorCsr.prototype._initCsr = function() {
 ripe.ConfiguratorCsr.prototype._deinitCsr = function() {
     this._destroyDebug();
     this._destroyInitialsResources();
+    this._destroyConfig(); // omit this as it should be called from configurator `deinit()`
```
NFSS10, line 1176: You forgot this

src/js/visual/configurator-csr.js
```diff
@@ -949,6 +955,13 @@ ripe.ConfiguratorCsr.prototype._destroyInitialsResources = function() {
     this._deinitCsrRenderedInitials();
 };
 
+/**
+ * Completely cleanup and destroy CSR config handlers.
+ */
+ripe.ConfiguratorCsr.prototype._destroyConfig = function() {
+    this._unregisterConfigGlobalHandlers();
+};
+
```
NFSS10, line 964: This doesn't make sense to exist as a `destroy` method as the context is registering and unregistering handlers.
Destroy methods (in this configurator context) are methods to completly self handle the release and destruction of a module of the configurator. Unregistering handlers is just one step of that.

The destroy method would make sense to exist if/when it is a self containing stuff. In this case the configurator `deinit()` is the so called `destroyConfig()` in this context

src/js/visual/configurator-csr.js
```diff
@@ -1759,7 +1774,7 @@ ripe.ConfiguratorCsr.prototype._onPostConfig = async function(self, config) {
 /**
  * @ignore
  */
-ripe.ConfiguratorCsr.prototype._registerHandlers = function() {
+ripe.ConfiguratorCsr.prototype._registerGlobalHandlers = function() {
```
NFSS10, line 1777: These first started as just the only existing handlers so the name made sense, but now the name is too general so if you are renaming it I I woul probably rename this to `_registerElementHandlers()`,  `_registerDomHandlers()` or something like that.

NFSS10 (changes requested):

src/js/visual/configurator-csr.js
3rdvision: This is a guard to fix the following problem:

When you have a configured model e.g a `dummy` `cube` and you call `ripe.config` to change to a different model e.g `dummy` `bill` there is a chain of events that's triggered. That chain of events includes the events `pre_config` `config` and `post_config`. Only after `post_config` is triggered you get to know if strategy `csr` is available for the configured build.

Since `_onPostConfig` was binded in a model that has CSR, when you change to a model that doesn't have CSR, `_onPostConfig` will get called and you have no way to prevent this from happening. 

Here's the debug log showing that CSR's `_onPostConfig` will happen before `_unregisterConfigGlobalHandlers` and we have no way of preventing this because the unmount of the CSR configurator takes more time than the `_onPostConfig`.

image

Users should be able to smoothly switch between different models but with the current CSR implementation this error will be thrown. I think `_onPostConfig` should not be running for strategies that don't include "CSR". In the current implementation `_onPostConfig` is running for any strategy after `ripe.config` so I think this guard should be in RIPE SDK and this problem should be dealt here.

NFSS10: I think it's working as intended then. If the model doesn't have CSR and you are trying to use CSR it should throw errors.
The one using this configurate is the one who should handle this, as in, it should deinit the configurator as it's not going to use it.
It's the same behaviour has having 3 thumbnails in a model and changing to a model that has only 2 thumbnails, you have to `deinit()` the extra one or it will error it out too.

Think like this, if you invert the roles and the norm was to have every build configurated to CSR instead of PRC and PRC were the ones that are only occasional configurated, then this change would be done there and not here which would not be correct to. A configurator is just another type of Visual component. It has the same weight as any other kind, as in, `configurator-prc`, `configurator-prc` and `image` have equal footing in the SDK.

3rdvision: > I think it's working as intended then.

Considering that we support switching models without having to do `configurator.deinit` how do you suggest for an implementation to do deal with this without throwing errors? If you wish you can share an example.html like before.

> Think like this, if you invert the roles and the norm was to have every build configurated to CSR instead of PRC and PRC were the ones that are only occasional configurated, then this change would be done there and not here

I think this change can be done in both places if we don't want to assume that there is always a PRC mode which I think is currently the case.
In the future we may have CSR only models and we might find more problems related to PRC not doing some deinits as it should.

NFSS10: We support changing to builds that have those capabilites. If you change to something unsupported it will throw erros. For example thumbnails.

As you can see, in White, the thumbnails destroy themselves when their number change for example:
https://github.com/ripe-tech/ripe-commons-pluginus/blob/f5fbdf5205ede56cfe5e5e7e8f19c00eb6cbab5a/vue/components/atoms/thumbnail/thumbnail.vue#L129

3rdvision: > the thumbnails destroy themselves when their number change

It's much lighter to destroy an image bind than to destroy a configurator bind so I don't think this is a fair comparison.

I think we should still support changing to builds that have different capabilities but given that white continues to work without this I'll resolve.

If a critical request comes up I'll bring this topic again.

src/js/visual/configurator-csr.js
```diff
@@ -1162,6 +1168,8 @@ ripe.ConfiguratorCsr.prototype._deinitCsr = function() {
     this._destroyInitialsResources();
     this._destroyScene();
 
+    this._unregisterConfigGlobalHandlers();
+
```
NFSS10, line 1172: Move this to `deinit()` here: https://github.com/ripe-tech/ripe-sdk/blob/563ba2394316f4ce2fdb6ecbed7f818afa246997/src/js/visual/configurator-csr.js#L157

As the registering is done in the `init()` thus the unregister should be done in it's corresponding `deinit()`

NFSS10 (changes requested):

NFSS10 (approved):

PR #500, opened 2023-03-08
#499 Load environment file bases on config

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/499 |
| Dependencies | -- |
| Decisions | - Stop using hardcoded Dropbox link (which is associated with my PlatformE account) and use PlatformE CDN<br>- Load environment file specified in the config |

### DIFF
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -142,14 +142,15 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
 /**
  * Returns the URL for the environment file of the 3D scene.
  *
- * @param {Object} options A map with options, such as:
- *  - 'brand' - The brand of the model.
- *  - 'model' - The name of the model.
- *  - 'version' - The version of the build, defaults to latest.
+ * @param {String} name The name of the environment map.
+ * @param {String} format The file format of the environment map.
+ * @param {Object} options A map with options.
  * @returns {String} The URL of the environment for the 3D scene.
  */
-ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
-    return "https://www.dl.dropboxusercontent.com/s/o0v07nn5egjrjl5/studio2.hdr";
+ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name, format, options) {
+    name = name || "studio2";
+    format = format || "hdr";
+    return `https://cdn.platforme.com/3d/environment/${name}.${format}`
 };
 
 /**
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -150,7 +150,7 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
 ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name, format, options) {
     name = name || "studio2";
     format = format || "hdr";
-    return `https://cdn.platforme.com/3d/environment/${name}.${format}`
+    return `https://cdn.platforme.com/3d/environment/${name}.${format}`;
 };
 
 /**
--- a/src/js/api/brand.js
+++ b/src/js/api/brand.js
@@ -147,9 +147,7 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
  * @param {Object} options A map with options.
  * @returns {String} The URL of the environment for the 3D scene.
  */
-ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name, format, options) {
-    name = name || "studio2";
-    format = format || "hdr";
+ripe.Ripe.prototype.getSceneEnvironmentUrl = function(name, format, options) {
     return `https://cdn.platforme.com/3d/environment/${name}.${format}`;
 };
 
--- a/src/js/visual/configurator-csr.js
+++ b/src/js/visual/configurator-csr.js
@@ -1309,6 +1309,9 @@ ripe.ConfiguratorCsr.prototype._unpackSceneOptions = function(options) {
     const cameraOptions = {};
     const zoomOptions = {};
 
+    // unpacks root options
+    const environment = options.environment;
+
     // unpacks renderer options
     rendererOptions.toneMapping = options.tone_mapping
         ? ripe.CsrUtils.toToneMappingValue(options.tone_mapping)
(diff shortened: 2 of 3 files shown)


src/js/api/brand.js
```diff
@@ -142,14 +142,15 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
 /**
  * Returns the URL for the environment file of the 3D scene.
  *
- * @param {Object} options A map with options, such as:
- *  - 'brand' - The brand of the model.
- *  - 'model' - The name of the model.
- *  - 'version' - The version of the build, defaults to latest.
+ * @param {String} name The name of the environment map.
+ * @param {String} format The file format of the environment map.
+ * @param {Object} options A map with options.
```
3rdvision, line 147:
```suggestion
 * @param {String} 'name' - The name of the environment map.

```

Use same format as others

src/js/api/brand.js
```diff
@@ -142,14 +142,15 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
...
+ * @param {String} name The name of the environment map.
+ * @param {String} format The file format of the environment map.
+ * @param {Object} options A map with options.
  * @returns {String} The URL of the environment for the 3D scene.
  */
-ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
-    return "https://www.dl.dropboxusercontent.com/s/o0v07nn5egjrjl5/studio2.hdr";
+ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name, format, options) {
+    name = name || "studio2";
+    format = format || "hdr";
```
3rdvision, line 152: why not
```suggestion
ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name = "studio2", format = "hdr", options) {

```

Also, our default hdr should be called studio instead of studio2
I think we should go as far as renaming the file in ripe_static

src/js/visual/configurator-csr.js
```diff
@@ -1309,6 +1309,9 @@ ripe.ConfiguratorCsr.prototype._unpackSceneOptions = function(options) {
     const cameraOptions = {};
     const zoomOptions = {};
 
+    // unpacks root options
+    const environment = options.environment;
+
```
3rdvision, line 1314: move this closer to the place where it's going to be used
or if you're ordering in the same order as the return then move the zoom block to be the last one

src/js/api/brand.js
```diff
@@ -142,14 +142,15 @@ ripe.Ripe.prototype.getFontUrl = function(name, format, options) {
...
- *  - 'model' - The name of the model.
- *  - 'version' - The version of the build, defaults to latest.
+ * @param {String} name The name of the environment map.
+ * @param {String} format The file format of the environment map.
+ * @param {Object} options A map with options.
  * @returns {String} The URL of the environment for the 3D scene.
  */
-ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(options) {
-    return "https://www.dl.dropboxusercontent.com/s/o0v07nn5egjrjl5/studio2.hdr";
+ripe.Ripe.prototype.get3dSceneEnvironmentUrl = function(name, format, options) {
```
3rdvision, line 150: I'd remove de 3d since it's implicit
```suggestion
ripe.Ripe.prototype.getSceneEnvironmentUrl = function(name, format, options) {

```

src/js/visual/configurator-csr.js
```diff
@@ -1373,8 +1377,9 @@ ripe.ConfiguratorCsr.prototype._loadCsrAssets = async function(
     this.dracoLoader = ripe.CsrUtils.loadDracoLoader();
 
     // computes environment file URL
-    const envUrl = this.owner.get3dSceneEnvironmentUrl();
-    const envFormat = "hdr";
+    const environment = sceneOptions.environment;
+    const environmentFormat = "hdr";
```
3rdvision, line 1381: I'd omit this since it's the default option

3rdvision (changes requested):
small things

src/js/api/brand.js
NFSS10: That's because it's the file given from 3D, I don't actually know what is the original name

src/js/visual/configurator-csr.js
NFSS10: It's not because I needed to define the format, it's because is a hardcoded value because this logic is not defined.

src/js/api/brand.js
NFSS10: The reason I don't put them in the arguments, is because the arguments are required and not optional. It's a fallback not a default

src/js/api/brand.js
NFSS10: The format is correct, check https://github.com/ripe-tech/ripe-sdk/blob/563ba2394316f4ce2fdb6ecbed7f818afa246997/src/js/api/config.js#L118

I think you are confusing with the documentation of object params.

src/js/visual/configurator-csr.js
NFSS10: This is a root value, it's following the same order as the other unpack methods

src/js/visual/configurator-csr.js
3rdvision: `loadEnvironment` defaults to 'hdr' right? https://github.com/ripe-tech/ripe-sdk/blob/563ba2394316f4ce2fdb6ecbed7f818afa246997/src/js/visual/csr/utils.js#L270

`get3dSceneEnvironmentUrl` fallsback to 'hdr' as well

src/js/api/brand.js
3rdvision: > It's a fallback not a default

Is there a difference in this case?

NFSS10: Yes! 😅 

Here it suggests the arguments are required. If you set like you suggestion, the method is letting you use those arguments are optional.

Some context for this type of usages:
The reason this fallback exists is for compatibility or it wouldn't even exist and just throw errors.
For example, a build that doesn't have the environment setting set, with this, will still work after this change (`dummy` `cube` for example) while still respecting the interface.

An alternative I guess would be to remove this from here and set it before it's usage. As I'm typing this now I see that this last approach is more correct so I will do that change now 🤔

src/js/visual/configurator-csr.js
NFSS10: This is here to enforce exactly to use `hdr`, it's not because it needs to pass it to `loadEnvironment()`, it's two different contexts

PR #507, opened 2023-04-12
feat: sku and domain in import order

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-core/issues/4798 |
| Dependencies | -- |
| Decisions | -- |
| Animated GIF | -- |

### DIFF
--- a/src/js/api/order.js
+++ b/src/js/api/order.js
@@ -2417,6 +2417,8 @@ ripe.Ripe.prototype._importOrder = function(ffOrderId, options = {}) {
     const quantity = options.quantity === undefined ? null : options.quantity;
     const brand = options.brand === undefined ? this.brand : options.brand;
     const factory = options.factory === undefined ? null : options.factory;
+    const domain = options.domain === undefined ? null : options.domain;
+    const sku = options.sku === undefined ? null : options.sku;
     const model = options.model === undefined ? this.model : options.model;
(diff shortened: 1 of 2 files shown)


src/js/api/order.js
```diff
@@ -2456,6 +2458,8 @@ ripe.Ripe.prototype._importOrder = function(ffOrderId, options = {}) {
     if (Object.keys(parts).length > 0) contents.parts = parts;
     if (size) contents.size = size;
     if (factory) contents.factory = factory;
+    if (domain) contents.domain = domain;
```
hugo-gomes, line 2461: try to sort the const declarations and ifs to be the same, meaning
if we're declaring `domain` and `sku` before `model`, lets add their if clauses before the `model`'s if clause

hugo-gomes (changes requested):

src/js/api/order.js
joao-conde: Likewise, we are declaring it after factory. The order is already wrong. I was trying not to change to many lines in this small PR

hugo-gomes: Can you fix the ordering on a separate PR please?

hugo-gomes (approved):

src/js/api/order.js
hugo-gomes: Thanks

Issue #511, opened 2023-05-19
Configurator ignoring size


joaofsl:
**Describe the bug**
The configurator ignores the `width` and `height` passed onto it.

**Impact**
This prevents the user from changing the default size of the images being rendered.

PR #512, opened 2023-05-19
#511 Configurator ignoring size

joaofsl:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-sdk/issues/511|
| Decisions | There is a boolean to force the sdk to use the `width` and  `height` to be used on the rendering call `useDefaultSize`. However that variable would always have a `true` value regardless of what was sent.|


3rdvision (approved):
Nice catch!

PR #517, opened 2023-06-21
#132 Custom compose url

3rdvision:
### Issue
- https://github.com/ripe-tech/products/issues/132

### Dependencies
- None

### Decisions
- Added `composeUrl` option to allow custom for composition endpoints

### Screenshots
- None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Added `sku` and `domain` fields to import order - [ripe-core/#4798](https://github.com/ripe-tech/ripe-core/issues/4798)
+* Added `composeUrl` option to allow custom for composition endpoints - [products/#132](https://github.com/ripe-tech/products/issues/97)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


hugo-gomes (changes requested):
@3rdvision add some unit tests

3rdvision:
> @3rdvision add some unit tests

An effective way of testing this would be to initialize ripe-sdk with a custom `composeUrl` and checking if the returned URL from `_getImageURL` would have the expected base URL.
Unfortunately, we don't do API testing in ripe-sdk so unless we change this high level decision, there's no space for a tests like these...

But this is indeed a risky change and it would be great if someone else could triple-check it.

NFSS10 (approved):

hugo-gomes (commented):
@3rdvision I was suggesting a test in the spirit of these ones https://github.com/ripe-tech/ripe-sdk/blob/7675715d925a28a7aa2e9aa244be4728092c8b84/test/js/base/api.js#L164

just as safety check, hence a nice to have
not mandatory at all

3rdvision:
@hugo-gomes Can you please re-approve please?
Fixed the changelog issue link.

hugo-gomes (approved):

